PR #12, opened 2016-11-27
Update crates.io version

GuillaumeGomez:


Cargo.toml
```diff
@@ -38,16 +38,16 @@ libc = "0.2"
...
 
 [dependencies.glib-sys]
 git = "https://github.com/gtk-rs/sys"
-version = "0.3.1"
+version = "0.3.2"
 
 [dependencies.glib]
 git = "https://github.com/gtk-rs/glib"
-version = "0.1.0"
+version = "0.3.2"
```
EPashkin: Seems wrong change

GuillaumeGomez: Arf, good catch once again!

Issue #31, opened 2017-06-05
Generate “open” signal for g_application


daa84:
I try to generate https://developer.gnome.org/gio/stable/GApplication.html#GApplication-open
When add Gio.File, Gir generate next closure:
```rust
Fn(&P, &[File], i32, &str)
```
But looks like this is wrong, because as i understand `&[File]` array can't be generated properly without n_files field, so looks like it is need to write this code manually? To make proper conversion between array of pointers to rust slice? Or any other way exists?
Thanks

GuillaumeGomez:
What is the `i32` for in here?

daa84:
i32 is n_files - count of files in [File] array
as i understand **GFile is not finished with null value

GuillaumeGomez:
Yes, that's my point. `n_files` is useless in Rust since it's supposed to be the length of the slice.

daa84:
Ok, but how to implement this properly? Gir generator does not support such kind of thing? Does it need to implement this manually? Or maybe some converter already exists?

GuillaumeGomez:
I wonder. I think we already did it but not sure if was in gir or manually...

cc @EPashkin

EPashkin:
Yes, this need be manual. Something like this:
```rust
impl<O: IsA<Application> + IsA<glib::object::Object>> ApplicationExtManual for O {
   fn connect_open<F: Fn(&Self, &[File], &str) + 'static>(&self, f: F) -> u64 {
        unsafe {
            let f: Box_<Box_<Fn(&Self, &[File], &str) + 'static>> = Box_::new(Box_::new(f));
            connect(self.to_glib_none().0, "open",
                transmute(open_trampoline::<Self> as usize), Box_::into_raw(f) as *mut _)
        }
    }
}
unsafe extern "C" fn open_trampoline<P>(this: *mut ffi::GApplication, files: *const *mut ffi::GFile, n_files: libc::c_int, hint: *mut libc::c_char, f: glib_ffi::gpointer)
where P: IsA<Application> {
    callback_guard!();
    let f: &Box_<Fn(&P, &[File], &str) + 'static> = transmute(f);
    let files: Vec<File> = FromGlibPtrContainer::from_glib_none_num(files, n_files as usize);
    f(&Application::from_glib_none(this).downcast_unchecked(), &files[..], &String::from_glib_none(hint))
}
```

daa84:
Thanks, i try 😄

PR #34, opened 2017-06-07
Implement open signal and open function

daa84:
Implement open signal and open function for #31


EPashkin:
Thanks, @daa84 
If both functions works, 👍

daa84:
Patch updated,
i checked "open" signal in my application and for me it works 😄

EPashkin:
My most worry is `open` functions. Maybe it need [&File] instead

daa84:
hm, yes, your right, this looks like error, i'll check

EPashkin:
Not necessarily an error. Just preparing an array of references IMHO can be easier.
However, the automatic `to_glib_none()` may work wrongly for this case.

daa84:
Ok, i check open method in application and it works without problem, also i check code for to_glib_none() in case of array it creates new temp array with call to_glib_none for every member so it looks like it must work 😄 
I think use [File] instead of [&File] is better because this method mirrors of "open" signal. It will be strange in case they are different.

EPashkin:
👍 
Thanks, @daa84.

PR #41, opened 2017-07-06
Add version for Notification::set_priority

EPashkin:
Part of https://github.com/gtk-rs/gir/issues/403


EPashkin:
cc @GuillaumeGomez. @jplatte

GuillaumeGomez:
It'd be better to be able to fully generate this but thanks a lot for the fix meanwhile!

jplatte:
Since I was cc'ed, thought I'd just mention that this PR does solve the issues I was having.

PR #47, opened 2017-07-31
Regen

GuillaumeGomez:
Apply gtk-rs/gir#419, gtk-rs/gir#418 and appveyour version fix

cc @EPashkin @sdroege


EPashkin:
👍 Thanks

PR #56, opened 2017-09-23
Update run

GuillaumeGomez:
It was incredibly annoying to use the `run` method when it was only taking `&str`s. With this, it's a lot easier.

cc @EPashkin @sdroege


sdroege:
Looks good to me

src/application.rs
```diff
@@ -12,10 +12,18 @@ use std::boxed::Box as Box_;
 use std::mem::transmute;
 
 pub trait ApplicationExtManual {
+    fn run(&self, argv: &[String]) -> i32;
     fn connect_open<F: Fn(&Self, &[File], &str) + 'static>(&self, f: F) -> SignalHandlerId;
 }
 
 impl<O: IsA<Application> + IsA<glib::object::Object>> ApplicationExtManual for O {
+    fn run(&self, argv: &[String]) -> i32 {
```
sdroege: I wonder if we could make this accept AsRef<str>, so that both &[String] and &[&str] could work

EPashkin: IMHO this possible only if we add `to_glib` for `&[AsRef(&str)]` beforehand

GuillaumeGomez: I did it at first but had a little lifetime issue (because I had to recreate a `Vec` with the references).

sdroege: I was thinking that it would be nice to be able to call `run(&["foo", "bar"])` without having to allocate memory. Also having run allow taking `std::env::Args` directly would seem useful.

Maybe this just needs a few separate functions instead of overgeneralizing :) Otherwise you could add a new `IntoArgs` trait...

GuillaumeGomez: Not sure that adding a new `IntoArgs` trait is such a good idea... But adding a `IntoEnv` trait seems better! (so like this, we implement it for the five types -three if we implement over `Vec<AsRef>` and `&[AsRef]`-).

PR #57, opened 2017-09-29
Dox

GuillaumeGomez:
cc @EPashkin

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -26,6 +26,7 @@ embed-lgpl-docs = ["gtk-rs-lgpl-docs"]
 v2_48 = ["v2_46", "gio-sys/v2_48"]
 v2_36 = ["v2_34", "gio-sys/v2_36"]
 v2_40 = ["v2_38", "gio-sys/v2_40"]
+dox = ["gio-sys/dox", "glib/dox"]
 
 [build-dependencies.gtk-rs-lgpl-docs]
 version = "0.1.3"


GuillaumeGomez:
Even though the regen is a bit useless. Should I remove it?

EPashkin:
Yes, seems only needed "Add doc feature"

GuillaumeGomez:
Only kept the last commit.

EPashkin:
Strange that gio features don't depends for glib's.
IMHO `dox` need depends of `glib/dox`

GuillaumeGomez:
Why that? If I follow your logic, `gtk` should add it for all its `gtk-rs`'s dependencies. Did I understood correctly?

EPashkin:
`gtk` can add only `gdk/v_11` only because `gdk` add features for its dependencies.
`gio` has only one non-sys dependency `glib` but not limit it,
so if any program use only `gio` with limited version, it may not get right version of `glib`'s functions.

Same if someone try to build docs only for `gio` he get all functions for it but only part for `glib`.

GuillaumeGomez:
Hum ok. Makes sense.

EPashkin:
As it very rare case I ignore version limits for glib as it will be added by other crates,
For `dox` I also don't insist but IMHO better add it.

GuillaumeGomez:
@EPashkin: Does it seem ok to you now?

EPashkin:
@GuillaumeGomez  Yes, thanks

PR #60, opened 2017-11-08
Create a copy of the Bytes passed to Resource::new_from_data() if it'…

sdroege:
…s not pointer-aligned

See https://bugzilla.gnome.org/show_bug.cgi?id=790030

Fixes https://github.com/gtk-rs/glib/issues/242

### DIFF
--- a/src/resource.rs
+++ b/src/resource.rs
@@ -1,0 +1,34 @@
+// Copyright 2017, The Gtk-rs Project Developers.
+// See the COPYRIGHT file at the top-level directory of this distribution.
+// Licensed under the MIT license, see the LICENSE file or <http://opensource.org/licenses/MIT>
+
+use Resource;
+use ffi;
+use glib_ffi;
+use glib;
+use glib::translate::*;
+use std::ptr;
+use std::mem;
+
+impl Resource {
+    pub fn new_from_data(data: &glib::Bytes) -> Result<Resource, glib::Error> {
+        unsafe {
+            let mut error = ptr::null_mut();
+
+            // Create a copy of data if it is not pointer-aligned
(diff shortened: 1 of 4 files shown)


src/resource.rs
```diff
@@ -0,0 +1,30 @@
...
+            // Create a copy of data if it is not pointer-aligned
+            // https://bugzilla.gnome.org/show_bug.cgi?id=790030
+            let mut data = data.clone();
+            let data_ptr = glib_ffi::g_bytes_get_data(data.to_glib_none().0, ptr::null_mut());
+            if data_ptr as usize % mem::align_of::<*const u8>() != 0 {
+                data = glib::Bytes::from(&*data);
+            }
+
+            let ret = ffi::g_resource_new_from_data(data.to_glib_none().0, &mut error);
+            if error.is_null() { Ok(from_glib_full(ret)) } else { Err(from_glib_full(error)) }
```
GuillaumeGomez: No one liner please. :)

GuillaumeGomez:
I followed the discussion on IRC and this seems to do what's expected. Just waiting for @EPashkin's confirmation.

src/resource.rs
sdroege: That's how it came from the autogenerated code (these kind of review comments that are wasting both our time are why I believe we should just use rustfmt and get over it...)

GuillaumeGomez: Thanks. And I told you that I wasn't completely opposed to rustfmt. If you have the patience and the time to set one up, then we'll use it.

EPashkin:
@sdroege Thanks.

@GuillaumeGomez  👍 from me

PR #63, opened 2017-11-23
Async code generation

antoyo:
Do not merge this PR.
It is only an example of the code generated by [this PR](https://github.com/gtk-rs/gir/pull/490) and I made this PR as requested by @sdroege.
This is generated using a newer version than the previous PR with the same name.


src/auto/app_info.rs
```diff
@@ -80,15 +84,26 @@ impl AppInfo {
...
-    //    unsafe { TODO: call ffi::g_app_info_launch_default_for_uri_async() }
-    //}
-
-    //#[cfg(any(feature = "v2_50", feature = "dox"))]
-    //pub fn launch_default_for_uri_finish<P: IsA</*Ignored*/AsyncResult>>(result: &P) -> Result<(), Error> {
-    //    unsafe { TODO: call ffi::g_app_info_launch_default_for_uri_finish() }
-    //}
+    #[cfg(any(feature = "v2_50", feature = "dox"))]
+    pub fn launch_default_for_uri_async<P: Fn(Result<(), Error>) + Send + Sync>(uri: &str, launch_context: &AppLaunchContext, callback: P) -> Cancellable {
+        let cancellable = Cancellable::new();
```
sdroege: There can be no Cancellable, or the same Cancellable could be used for multiple operations. Maybe let the user allocate it if needed (it's not cheap) and let the user pass an `Option<&Cancellable>`?

src/auto/app_info.rs
```diff
@@ -80,15 +84,26 @@ impl AppInfo {
...
+        let user_data: Box<Box<Fn(Result<(), Error>) + Send + Sync>> = Box::new(Box::new(callback));
+        extern "C" fn launch_default_for_uri_async_trampoline(_source_object: *mut gobject_ffi::GObject, res: *mut ffi::GAsyncResult, user_data: glib_ffi::gpointer)
+        {
+            unsafe {
+                let mut error = ptr::null_mut();
+                let _ = ffi::g_app_info_launch_default_for_uri_finish(res, &mut error);
+                let result = if error.is_null() { Ok(()) } else { Err(from_glib_full(error)) };
+                let callback: &Box<Fn(Result<(), Error>) + Send + Sync> = transmute(user_data);
```
sdroege: clippy will warn about this one. First it could be &&Fn (that's easy to fix), second it could be a cast instead of transmute (that's very noisy and we didn't fix it for signals either).

Also I like that the trampoline is inside the function. We should do the same for signals (can you create a GitHub ticket about that?).

sdroege:
Is there a reason why all the async functions on GFile are not generated? Are they not supported yet for one reason or another?

src/auto/app_info.rs
sdroege: Or rather, the signal trampolines should disappear the way they are and replaced by a Closure callback :)

antoyo:
@sdroege Sometimes, the async methods are not generated for `GFile` because a type is missing, this type could be the `ProgressCallback` as we discussed in Berlin.

src/auto/app_info.rs
antoyo: By "There can be no Cancellable", you mean the user could want no `Cancellable` and you want to have the same API as in C, meaning that the user decide whether there is a `Cancellable` or not?

sdroege:
`g_file_open_readwrite_async()` for example should have no unknown types but does not appear here

src/auto/app_info.rs
sdroege: Yes. The user could create a `Cancellable` or pass `None`. And the user could use the same `Cancellable` for multiple operations, even simultaneously or one after another.

sdroege:
Oh, oops :) Nevermind then

sdroege:
Seems good, thanks

src/auto/app_info.rs
```diff
@@ -80,15 +84,26 @@ impl AppInfo {
...
+        let cancellable = cancellable.into();
+        let cancellable = cancellable.to_glib_none();
+        let user_data: Box<Box<Fn(Result<(), Error>) + Send + Sync>> = Box::new(Box::new(callback));
+        extern "C" fn launch_default_for_uri_async_trampoline(_source_object: *mut gobject_ffi::GObject, res: *mut ffi::GAsyncResult, user_data: glib_ffi::gpointer)
+        {
+            unsafe {
+                let mut error = ptr::null_mut();
+                let _ = ffi::g_app_info_launch_default_for_uri_finish(res, &mut error);
+                let result = if error.is_null() { Ok(()) } else { Err(from_glib_full(error)) };
+                let callback: &&Fn(Result<(), Error>) = transmute(user_data);
```
sdroege: This technically still misses the Send/Sync bounds, but doesn't really matter

src/auto/app_info.rs
```diff
@@ -80,15 +84,26 @@ impl AppInfo {
...
-    //pub fn launch_default_for_uri_async<'a, 'b, P: Into<Option<&'a Cancellable>>, Q: Into<Option<&'b /*Unimplemented*/AsyncReadyCallback>>, R: Into<Option</*Unimplemented*/Fundamental: Pointer>>>(uri: &str, launch_context: &AppLaunchContext, cancellable: P, callback: Q, user_data: R) {
-    //    unsafe { TODO: call ffi::g_app_info_launch_default_for_uri_async() }
-    //}
-
-    //#[cfg(any(feature = "v2_50", feature = "dox"))]
-    //pub fn launch_default_for_uri_finish<P: IsA</*Ignored*/AsyncResult>>(result: &P) -> Result<(), Error> {
-    //    unsafe { TODO: call ffi::g_app_info_launch_default_for_uri_finish() }
-    //}
+    #[cfg(any(feature = "v2_50", feature = "dox"))]
+    pub fn launch_default_for_uri_async<'a, P: Into<Option<&'a Cancellable>>, Q: Fn(Result<(), Error>) + Send + Sync>(uri: &str, launch_context: &AppLaunchContext, cancellable: P, callback: Q) {
```
sdroege: The closures must all also be `'static`

sdroege:
👍

GuillaumeGomez:
Ready to be merged then?

EPashkin:
It need regen with last gir

GuillaumeGomez:
@antoyo: Can you regen please so we can move forward?

antoyo:
@GuillaumeGomez Done.

EPashkin:
@antoyo Please update gir submodule too

antoyo:
@EPashkin Done.

EPashkin:
@antoyo Thanks.

GuillaumeGomez:
Thanks!

GuillaumeGomez:
@EPashkin: Are we ready to move now? :)

EPashkin:
@GuillaumeGomez  IMHO yes

GuillaumeGomez:
Then let's go!

PR #69, opened 2017-12-09
Generate streams

EPashkin:
cc @GuillaumeGomez, @sdroege


GuillaumeGomez:
👍

Gir.toml
```diff
@@ -72,6 +94,14 @@ status = "generate"
     name = "run"
     ignore = true
 
+[[object]]
+name = "Gio.DataInputStream"
+status = "generate"
+    [[object.function]]
+    name = "read_line"
+    #return vec of char*
+    ignore = true
```
sdroege: That should just return a `Vec<u8>` and I thought gir handles that by now (note that the element type is set to guint8 in the gir file, it's not utf8).

Gir.toml
```diff
@@ -95,6 +125,14 @@ status = "generate"
     name = "hash"
     ignore = true
 
+[[object]]
+name = "Gio.IOStream"
+status = "generate"
+    [[object.function]]
+    name = "splice_async"
+    #g_io_stream_splice_finish accept only 2 parameters instead normal 3 (no source_object)
+    ignore = true
```
sdroege: @antoyo ↑

src/auto/buffered_input_stream.rs
```diff
@@ -0,0 +1,153 @@
...
+pub trait BufferedInputStreamExt {
+    fn fill<'a, P: Into<Option<&'a Cancellable>>>(&self, count: isize, cancellable: P) -> Result<isize, Error>;
+
+    fn fill_async<'a, P: Into<Option<&'a Cancellable>>, Q: Fn(Result<(), Error>) + Send + Sync + 'static>(&self, count: isize, io_priority: i32, cancellable: P, callback: Q);
+
+    fn get_available(&self) -> usize;
+
+    fn get_buffer_size(&self) -> usize;
+
+    fn peek(&self, buffer: &[u8], offset: usize) -> usize;
```
sdroege: The buffer needs to be mutable, the function reads into this buffer

src/auto/converter.rs
```diff
@@ -0,0 +1,47 @@
...
+}
+
+pub trait ConverterExt {
+    fn convert(&self, inbuf: &[u8], outbuf: &[u8], flags: ConverterFlags) -> Result<(ConverterResult, usize, usize), Error>;
+
+    fn reset(&self);
+}
+
+impl<O: IsA<Converter>> ConverterExt for O {
+    fn convert(&self, inbuf: &[u8], outbuf: &[u8], flags: ConverterFlags) -> Result<(ConverterResult, usize, usize), Error> {
```
sdroege: outbuf must be mutable

src/auto/functions.rs
```diff
@@ -300,20 +312,42 @@ pub fn networking_init() {
 //    unsafe { TODO: call ffi::g_pollable_source_new_full() }
 //}
 
-//#[cfg(any(feature = "v2_34", feature = "dox"))]
-//pub fn pollable_stream_read<'a, P: IsA</*Ignored*/InputStream>, Q: Into<Option<&'a Cancellable>>>(stream: &P, buffer: &[u8], blocking: bool, cancellable: Q) -> Result<isize, Error> {
-//    unsafe { TODO: call ffi::g_pollable_stream_read() }
-//}
+#[cfg(any(feature = "v2_34", feature = "dox"))]
+pub fn pollable_stream_read<'a, P: IsA<InputStream>, Q: Into<Option<&'a Cancellable>>>(stream: &P, buffer: &[u8], blocking: bool, cancellable: Q) -> Result<isize, Error> {
```
sdroege: This should not be in functions, but in InputStream (and the other one in OutputStream). Also calling this with blocking=false on a non-GPollableInputStream is undefined behaviour, thus unsafe. I would just not expose this at all, it's C convenience stuff with footguns

sdroege: Also the buffer needs to be writable

src/auto/input_stream.rs
```diff
@@ -0,0 +1,248 @@
...
+    fn is_closed(&self) -> bool;
+
+    fn read<'a, P: Into<Option<&'a Cancellable>>>(&self, buffer: &[u8], cancellable: P) -> Result<isize, Error>;
+
+    fn read_all<'a, P: Into<Option<&'a Cancellable>>>(&self, buffer: &[u8], cancellable: P) -> Result<usize, Error>;
+
+    #[cfg(any(feature = "v2_44", feature = "dox"))]
+    fn read_all_async<'a, P: Into<Option<&'a Cancellable>>, Q: Fn(Result<usize, Error>) + Send + Sync + 'static>(&self, buffer: &[u8], io_priority: i32, cancellable: P, callback: Q);
+
+    fn read_async<'a, P: Into<Option<&'a Cancellable>>, Q: Fn(Result<(), Error>) + Send + Sync + 'static>(&self, buffer: &[u8], io_priority: i32, cancellable: P, callback: Q);
```
sdroege: Buffer in all the read functions must be mutable

src/auto/input_stream.rs
```diff
@@ -0,0 +1,248 @@
...
+
+    fn close<'a, P: Into<Option<&'a Cancellable>>>(&self, cancellable: P) -> Result<(), Error>;
+
+    fn close_async<'a, P: Into<Option<&'a Cancellable>>, Q: Fn(Result<(), Error>) + Send + Sync + 'static>(&self, io_priority: i32, cancellable: P, callback: Q);
+
+    fn has_pending(&self) -> bool;
+
+    fn is_closed(&self) -> bool;
+
+    fn read<'a, P: Into<Option<&'a Cancellable>>>(&self, buffer: &[u8], cancellable: P) -> Result<isize, Error>;
```
sdroege: Buffer in all the read functions must be mutable

src/auto/input_stream.rs
sdroege: And this buffer must even stay alive until the async callback is called

src/auto/output_stream.rs
```diff
@@ -0,0 +1,304 @@
...
+    fn flush_async<'a, P: Into<Option<&'a Cancellable>>, Q: Fn(Result<(), Error>) + Send + Sync + 'static>(&self, io_priority: i32, cancellable: P, callback: Q);
+
+    fn has_pending(&self) -> bool;
+
+    fn is_closed(&self) -> bool;
+
+    fn is_closing(&self) -> bool;
+
+    //#[cfg(any(feature = "v2_40", feature = "dox"))]
+    //fn printf<'a, P: Into<Option<&'a Cancellable>>>(&self, cancellable: P, error: &mut Error, format: &str, : /*Unknown conversion*//*Unimplemented*/Fundamental: VarArgs) -> Option<usize>;
```
sdroege: Wat! I didn't know this function exists. Good to know :)

src/auto/output_stream.rs
```diff
@@ -0,0 +1,304 @@
...
+    //fn vprintf<'a, P: Into<Option<&'a Cancellable>>>(&self, cancellable: P, error: &mut Error, format: &str, args: /*Unknown conversion*//*Unimplemented*/Unsupported) -> Option<usize>;
+
+    fn write<'a, P: Into<Option<&'a Cancellable>>>(&self, buffer: &[u8], cancellable: P) -> Result<isize, Error>;
+
+    fn write_all<'a, P: Into<Option<&'a Cancellable>>>(&self, buffer: &[u8], cancellable: P) -> Result<usize, Error>;
+
+    #[cfg(any(feature = "v2_44", feature = "dox"))]
+    fn write_all_async<'a, P: Into<Option<&'a Cancellable>>, Q: Fn(Result<usize, Error>) + Send + Sync + 'static>(&self, buffer: &[u8], io_priority: i32, cancellable: P, callback: Q);
+
+    fn write_async<'a, P: Into<Option<&'a Cancellable>>, Q: Fn(Result<(), Error>) + Send + Sync + 'static>(&self, buffer: &[u8], io_priority: i32, cancellable: P, callback: Q);
```
sdroege: Buffer must stay alive until the async callback (or rather the finish method) is called

Gir.toml
EPashkin: Gir generated good looking code:
```rust
    fn read_line<'a, P: Into<Option<&'a Cancellable>>>(&self, cancellable: P) -> Result<(Vec<u8>, usize), Error> {
        let cancellable = cancellable.into();
        let cancellable = cancellable.to_glib_none();
        unsafe {
            let mut length = mem::uninitialized();
            let mut error = ptr::null_mut();
            let ret = ffi::g_data_input_stream_read_line(self.to_glib_none().0, &mut length, cancellable.0, &mut error);
            if error.is_null() { Ok((FromGlibPtrContainer::from_glib_full(ret), length)) } else { Err(from_glib_full(error)) }
        }
    }
```
but compiling get 2 errors:
```
error[E0277]: the trait bound `u8: glib::translate::Ptr` is not satisfied
   --> src\auto\data_input_stream.rs:145:38
    |
145 |             if error.is_null() { Ok((FromGlibPtrContainer::from_glib_full(ret), length)) } else { Err(from_glib_full(error)) }
    |                                      ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^ the trait `glib::translate::Ptr` is not implemented for `u8`
    |
    = note: required because of the requirements on the impl of `glib::translate::FromGlibPtrContainer<u8, *mut u8>` for `std::vec::Vec<u8>`
    = note: required by `glib::translate::FromGlibPtrContainer::from_glib_full`

error[E0277]: the trait bound `u8: glib::translate::FromGlibPtrArrayContainerAsVec<u8, *mut u8>` is not satisfied
   --> src\auto\data_input_stream.rs:145:38
    |
145 |             if error.is_null() { Ok((FromGlibPtrContainer::from_glib_full(ret), length)) } else { Err(from_glib_full(error)) }
    |                                      ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^ the trait `glib::translate::FromGlibPtrArrayContainerAsVec<u8,
 *mut u8>` is not implemented for `u8`
    |
    = note: required because of the requirements on the impl of `glib::translate::FromGlibPtrContainer<u8, *mut u8>` for `std::vec::Vec<u8>`
    = note: required by `glib::translate::FromGlibPtrContainer::from_glib_full`
```

sdroege: Hmm, possibly a bug in glib/translate.rs. Exactly that impl should be used here and should be usable, unclear why it isn't. The `glib::translate::Ptr` requirement looks dubious.

Can you create a ticket?

sdroege:
Mostly looks ok, just that all read functions need a mutable buffer, all read/write/etc async functions need their buffer to stay alive until the async callback is called, and various other minor things.

Gir.toml
EPashkin: Added https://github.com/gtk-rs/glib/issues/275

src/auto/buffered_input_stream.rs
EPashkin: Seems this function need be manual :(

src/auto/converter.rs
EPashkin: Will made all Converter manual

src/auto/input_stream.rs
EPashkin: There way to do it?

sdroege: With autogenerated code? No, not right now

src/auto/functions.rs
EPashkin: Ok I ignore this function.

src/auto/input_stream.rs
EPashkin: In manual code?

src/auto/functions.rs
EPashkin: Maybe ignore `pollable_stream_write` too? PollableStreams not generated.

src/auto/input_stream.rs
```diff
@@ -0,0 +1,170 @@
...
+                let result = if error.is_null() { Ok(()) } else { Err(from_glib_full(error)) };
+                let callback: &&(Fn(Result<(), Error>) + Send + Sync + 'static) = transmute(user_data);
+                callback(result);
+            }
+        }
+        let callback = read_bytes_async_trampoline;
+        unsafe {
+            ffi::g_input_stream_read_bytes_async(self.to_glib_none().0, count, io_priority, cancellable.0, Some(callback), Box::into_raw(user_data) as *mut _);
+        }
+    }
```
EPashkin: Seems this function generated wrong, see https://developer.gnome.org/gio/stable/GInputStream.html#g-input-stream-read-bytes-async
Don't understand how user get `glib::Bytes` from it 
cc @antoyo

sdroege: g_input_stream_read_bytes_finish() gives you the GBytes

src/auto/functions.rs
sdroege: Yes, I meant both

src/auto/input_stream.rs
sdroege: Yes

EPashkin:
Made `read_bytes_async` manual, 
There still question, how to "buffer must even stay alive until the async callback is called" even in manual code and this related to almost all async functions :(

EPashkin:
Make some async functions manual, moved entire `file_input_stream.rs` to manual as it have only two function and one of its `query_info_async` - wrong.
Will add other changes as separate commits, to simplify checking.

Gir.toml
antoyo: Do I need to update `gir` or something?

antoyo:
@EPashkin So, if I understand correctly, what's missing is the return value (`Ok` value of the `Result`), right?

EPashkin:
Yes.
In `Result`-returning functions we normally skip it only if return type `void` or `bool`

EPashkin:
@sdroege All async IMHO "fixed"

EPashkin:
But forgot make some buffers mutable, still fixing :(

EPashkin:
Done too.
Only last problem is "buffer must even stay alive until the async callback is called" for all async operation.

sdroege:
> There still question, how to "buffer must even stay alive until the async callback is called" even in manual code and this related to almost all async functions :(

Look at the GStreamer Buffer function I linked elsewhere: https://github.com/sdroege/gstreamer-rs/blob/963557b79f5d3a91db944ed953a8c9fcab78a6ba/gstreamer/src/buffer.rs#L100

EPashkin:
Sorry, I don't understand.

sdroege:
You can take any `AsMut<[u8]>` here, but let's ignore that part for now and take `Vec<u8>`. You need to box it in the async function and then make that part of the `user_data` (https://github.com/sdroege/gstreamer-rs/blob/963557b79f5d3a91db944ed953a8c9fcab78a6ba/gstreamer/src/buffer.rs#L76-L81). In the finish function you would steal it from the `user_data` (i.e. move it out of there!), in the destroy_notify for the user_data you would drop it (like https://github.com/sdroege/gstreamer-rs/blob/963557b79f5d3a91db944ed953a8c9fcab78a6ba/gstreamer/src/buffer.rs#L67) if it was not stolen (see before).

EPashkin:
Is I not wrong on you just save pointer and size without blocking owner of memory do what his want. 
So I don't see how this protect this memory block from free and pollute with other data.

sdroege:
> Is I not wrong on you just save pointer and size without blocking owner of memory do what his want.
So I don't see how this protect this memory block from free and pollute with other data.

The function takes ownership of the data passed in, and only ever releases it from the destroy_notify or when the user releases it after calling the finish function.

EPashkin:
So you meant that [`write_async`](https://github.com/gtk-rs/gio/pull/69/files#diff-7c71e04c0e2366e61ffa11af9cc3e345R49) must take `buffer: Vec<u8>` instead `buffer: &[u8]`,
and always drop it in `write_async_trampoline` ?

sdroege:
Yes, look at the code I linked above. Either `Vec<u8>` or `AsRef<[u8]>`.

EPashkin:
`AsRef<[u8]>` IMHO don't transfer ownership inside function.

sdroege:
> AsRef<[u8]> IMHO don't transfer ownership inside function.

(I was omitting some trait bounds for simplicity, see the code I linked for what exactly you need)
`fn bla<T: AsRef<[u8]> + Send + 'static>(t: T)` does transfer ownership, `fn bla<T: AsRef<[u8]> + Send + 'static>(t: &T)` doesn't.

EPashkin:
Oh, I missed `b:&T` vs `b:T` :(
Still without `+'static`, rust allow this:
```rust
    fn tst(&self) {
        let mut buf = vec![1u8,2,3,4,5,7];
        self.write_async(&buf, 1, None, callback);
        buf[0]=2;
    }
    fn write_async<'a, T: AsRef<[u8]> + Send /*+ 'static*/, P: Into<Option<&'a Cancellable>>, Q: Fn(Result<isize, Error>) + Send + Sync + 'static>(&self, buffer: T, io_priority: i32, cancellable: P, callback: Q) {
```

With `+'static` it don't need drop, but IMHO still allowed.
Get `internal compiler error` on trying repeat example :(
Can you show working example of `AsRef<[u8]> + 'static` usage (actual version of your file have `pub fn from_vec(vec: Vec<u8>) -> Option<Self> {` ) ?

EPashkin:
I think that for async operations with buffer it need be returned in callback.
I get this compiled as example, but not tested if it works.
This ugly :(
```
    fn write_async<'a, P: Into<Option<&'a Cancellable>>, Q: Fn(Vec<u8>, Result<isize, Error>) + Send + Sync + 'static>(&self, buffer: Vec<u8>, io_priority: i32, cancellable: P, callback: Q) {
        let cancellable = cancellable.into();
        let cancellable = cancellable.to_glib_none();
        let count = buffer.len() as usize;
        let mut user_data: Box<(Box<Fn(Vec<u8>, Result<isize, Error>) + Send + Sync + 'static>, Vec<u8>)> = Box::new((Box::new(callback), buffer));
        let ptr = user_data.1.as_mut_ptr();
        extern "C" fn write_async_trampoline(_source_object: *mut gobject_ffi::GObject, res: *mut ffi::GAsyncResult, user_data: glib_ffi::gpointer)
        {
            callback_guard!();
            unsafe {
                let mut error = ptr::null_mut();
                let ret = ffi::g_output_stream_write_finish(_source_object as *mut _, res, &mut error);
                let result = if error.is_null() { Ok(ret) } else { Err(from_glib_full(error)) };
                let box_: Box<(Box<Fn(Vec<u8>, Result<isize, Error>) + Send + Sync + 'static>, Vec<u8>)> = Box::from_raw(transmute(user_data));
                let box_ = *box_;
                let (cb, buf) = box_;
                cb(buf, result);
            }
        }
        let callback = write_async_trampoline;
        unsafe {
            ffi::g_output_stream_write_async(self.to_glib_none().0, ptr, count, io_priority, cancellable.0, Some(callback), Box::into_raw(user_data) as *mut _);
        }
    }

```

sdroege:
As I said above, and also do in the linked code, you need `'static` (and `Send`). Otherwise exactly the problem you mention is possible, and that's not good :)

sdroege:
Why do you think it must be returned in callback? It can, but it does not have to. You can as well drop it from the trampoline (and in case of the read functions, you get it back from the finish() function and that you indeed want to pass to the callback).

sdroege:
Your code should work though, but you could also use `AsRef<[u8]>` here. Then for example the `Chunk` type from hyper could also be directly passed here instead of copying to a `Vec` first.

EPashkin:
`+'static` still allow pass reference, so `AsRef` is not safe anyway :(
```
static buf0:&[u8] = &[1,2,3];
fn callback(ret: Result<isize, Error>) {}
//in trait
fn tst(&self) {
        self.write_async(&buf0, 1, None, callback);
        let b = buf0[0];//=2;
    }
    fn write_async<'a, T: AsRef<[u8]> + Send + 'static, P: Into<Option<&'a Cancellable>>, Q: Fn(Result<isize, Error>) + Send + Sync + 'static>(&self, buffer: T, io_priority: i32, cancellable: P, callback: Q) {
```

Also it seems don't work good with `&mut [u8]` (but at minimum it prevent reference passing),
so readers can't be done this way.

Maybe better way add some type, that works like `Arc<RefCell<T>>` but allow block access while async action in progress.

EPashkin:
Forgot to say that [`write_bytes_async`](https://github.com/gtk-rs/gio/pull/69/files#diff-7c71e04c0e2366e61ffa11af9cc3e345R71) will need "lock" `glib::Bytes` and `+'static` IMHO don't make sense for this case.

sdroege:
I'll write you an async read and write function tomorrow, then we have something to discuss about :)

sdroege:
Also generally, you *must* not use `&[u8]` for the async read/write functions, unless they have a `'static` lifetime. Preparing something now :)

sdroege:
@EPashkin https://github.com/sdroege/gio-rs/commit/10dd01f30809cc72f768dc70bb3c8bed9b4e3f88 This is what I meant, see the commit message for the details of all changes (I also noticed a bug with all async methods currently :( see https://github.com/gtk-rs/gir/issues/508 ).

sdroege:
Tests for this can be easily written around the memory input/output streams btw

EPashkin:
Easy for async functions? 😉

EPashkin:
There was [build fix in `read_all_async`](https://github.com/gtk-rs/gio/commit/8444a1378a7106a361bacd532103cf13bfcfd8e3)

Some changes in `read_all_async` that I propose:
[don't modify buffer but return readed instead](https://github.com/gtk-rs/gio/commit/ce88681a63fd5504dd28cc047df5c7876a1e4a0f)
[Use `AsMut<[u8]>` instead array](https://github.com/gtk-rs/gio/commit/ae24af418803168f77c4ee06cea6c5c271faa4ed)

EPashkin:
@sdroege Updated `read_all_async` to use `Fn(Vec<u8>, Option<Error>)`

sdroege:
Seems ok now, will go over it another time in detail later/tomorrow

EPashkin:
Ok, so this question is for later.
We will use `AsMut<[u8]>` for this function?
What we pass to clouse then: `AsMut<[u8]>` or actual type?

sdroege:
For which function?

EPashkin:
Now only for `read_all_async`

sdroege:
The read ones must be Vec as you need set_len

EPashkin:
You right, it need for all read functions. If we don't want add special trait for this 😉

EPashkin:
Changed `InputStream::read_all`, maybe error tuple should reversed?
Added tests for `MemoryInputStream`.
Found that `new_from_data` and `add_data` can't be done in safe way,
as internally `gio` creates `glib::Bytes` from buffer with passing same pointer to `destroy` function,
so way used in `Bytes::from_owned` impossible.
@sdroege can you check this when you have free time?

sdroege:
Yes, only the functions that take a GBytes work with bindings (unless you copy into new malloc'd memory or similar but don't)

EPashkin:
Then maybe add version "v2_34" for `MemoryInputStream` to limit it usage?
And for `MemoryOutputStream` it will be "v2_36"

sdroege:
I guess you could in theory get such a stream from other API and only use the generic stream API on it

EPashkin:
Ok.

EPashkin:
@GuillaumeGomez Please restart Travis here.

GuillaumeGomez:
Done.

sdroege:
Might want to wait for https://github.com/gtk-rs/gir/pull/509 (it also affects the manual code, that would need some updating)

src/lib.rs
```diff
@@ -29,13 +38,26 @@ pub use glib::{
 pub use auto::*;
 pub use auto::functions::*;
 
+use converter::Converter;
+use file_i_o_stream::FileIOStream;
+use file_input_stream::FileInputStream;
+use file_output_stream::FileOutputStream;
```
sdroege: Why not *?

src/input_stream.rs
```diff
@@ -0,0 +1,180 @@
...
+use std::ptr;
+use InputStream;
+
+pub trait InputStreamExtManual {
+    fn read<'a, B: AsMut<[u8]>, P: Into<Option<&'a Cancellable>>>(&self, buffer: B, cancellable: P) -> Result<isize, Error>;
+
+    fn read_all<'a, P: Into<Option<&'a Cancellable>>>(&self, buffer: Vec<u8>, cancellable: P) -> Result<Vec<u8>, (Error, Vec<u8>)>;
+
+    #[cfg(any(feature = "v2_44", feature = "dox"))]
+    fn read_all_async<'a, P: Into<Option<&'a Cancellable>>, Q: Fn(Vec<u8>, Option<Error>) + Send + 'static>(&self, buffer: Vec<u8>, io_priority: i32, cancellable: P, callback: Q);
```
sdroege: @antoyo might have an opinion on this one for the futures-glib stuff. I didn't look at that, but I assume you require the closures to get a `Result` passed, which is then converted into `Future::Item` and `Future::Error`? That might be an argument for making this a `Result<Vec<u8>, (Vec<u8>, Error)>` instead.

And generally, having a common "shape" for the async functions could be beneficial in the future for some generic higher-level thing around them.

sdroege:
Maybe also an async read test for the memory input stream could be useful. basically you could do it similar to the MainContext::invoke() test: create a main loop, "schedule" the read (i.e. call read_all_async()) and from the closures check the read data and quit the main loop. And after the call to read_all_async() you would run the main loop.

EPashkin:
Thanks, I was no idea how to do async test, and planned to look gtk/glib source to get some hints.

EPashkin:
@sdroege, @GuillaumeGomez I have question about `MemoryOutputStream`.
There one function that IMHO need for this is [`get_data`](https://developer.gnome.org/gio/stable/GMemoryOutputStream.html#g-memory-output-stream-get-data).
We can done just `Deref<Target=[u8]>` for stream,
but I prefer add `with_data(FnOnce(data: &[u8]))` instead, as data "may become invalid on the next write or truncate operation on the stream"
This can add problem for usage but add safety.
Maybe there other way?

GuillaumeGomez:
Can't we just have a specific type which will handle with corresponding lifetime?

sdroege:
The problem is that the memory output stream is reference counted. You could just have two references and then unsafety happens.

sdroege:
`get_data` needs to copy, everything else is unsafe. `steal_data/bytes` is fine and does not need to copy

src/input_stream.rs
EPashkin: Minor question here what better `(Error, Vec<u8>)` or `(Vec<u8>, Error)` ?

sdroege: Hm... or maybe even `Result<(Vec<u8>, Option<Error>), Error>`? The error case with read bytes is not *really* an error

EPashkin:
Problem with `steal_data` is "when no longer needed it must be freed using the free function set in ostream 's “destroy-function” property".
So I think better ignore it and  and use `steal_as_bytes` instead.

EPashkin:
Only one problem with `steal_as_bytes` that it one time operation, so it won't allow "peek" stream between writes.

src/input_stream.rs
EPashkin: Then when we return `Err(err)` part?

sdroege:
You could add a `fn peek_data<F: FnOnce(&[u8])>(&self, func: F)` that calls `get_data()` and uses `slice::from_raw_parts()`

EPashkin:
Yes, is that I meant by my `with_data`, but now I don't sure that it not invalidated while closure called.

sdroege:
Ah indeed, you could pass a (second) reference of the stream into the closure and then cause unsafety.

sdroege:
Only `steal_as_bytes` it is then :)

src/input_stream.rs
sdroege: When 0 bytes were read and there was an error

EPashkin: IMHO we need return buffer in all error cases  to reuse be possible.
So `Result<Vec<u8>, (Vec<u8>, Error)>` or with reversed error tuple is only options.

sdroege: `Result<(Vec<u8>, Option<Error>), (Vec<u8>, Error)>`?

EPashkin: Maybe. So `read_all` need same change

EPashkin:
@sdroege Added tests for `read_all_async` with old `FnOnce(Vec<u8>, Option<Error>)` will change for `Result<(Vec<u8>, Option<Error>), (Vec<u8>, Error)>` later.
Take look please.

EPashkin:
There some results: 
1) async functions with `io_priority` need be `glib::Priority` instead `i32`.
2) Streams need at minimum `Send`, not sure that it right.

sdroege:
1) yes
2) why do you need them Send? But yes, and that's unsafe :( they are Send but not Sync, but they can't be Send either because reference counting

EPashkin:
1. Ok, I start replace it.
2. Thought use in tests but seems this unneeded

src/input_stream.rs
```diff
@@ -0,0 +1,270 @@
...
+
+            callback(result);
+        }
+        let callback = read_all_async_trampoline::<Q>;
+        unsafe {
+            ffi::g_input_stream_read_all_async(self.to_glib_none().0, buffer_ptr, capacity, io_priority.to_glib(), cancellable.0, Some(callback), Box::into_raw(user_data) as *mut _);
+        }
+    }
+
+    fn read_async<'a, P: Into<Option<&'a Cancellable>>, Q: FnOnce(Result<Vec<u8>, (Vec<u8>, Error)>) + Send + 'static>(&self, mut buffer: Vec<u8>, io_priority: Priority, cancellable: P, callback: Q) {
```
EPashkin: @sdroege Can you check this function, tests with it failing with AV.
Seems I made some mistake, but can't find it.

src/input_stream.rs
```diff
@@ -0,0 +1,270 @@
...
+        let cancellable = cancellable.into();
+        let cancellable = cancellable.to_glib_none();
+        let buffer_ptr = buffer.as_mut_ptr();
+        let capacity = buffer.capacity() as usize;
+        let user_data: Box<Option<(Box<Q>, Vec<u8>)>> = Box::new(Some((Box::new(callback), buffer)));
+        unsafe extern "C" fn read_async_trampoline<Q: FnOnce(Result<Vec<u8>, (Vec<u8>, Error)>) + Send + 'static>(_source_object: *mut gobject_ffi::GObject, res: *mut ffi::GAsyncResult, user_data: glib_ffi::gpointer)
+        {
+            callback_guard!();
+            let mut user_data: Box<Option<(Q, Vec<u8>)>> = Box::from_raw(user_data as *mut _);
```
EPashkin: Sorry, found that i forgot `Box` over `Q`

sdroege:
All good and working now and just needs another review, or was there an open question?

EPashkin:
@sdroege Can you look to `src/inputstream.rs`, please, when you have time? Maybe something in it can be done better.

sdroege:
Seems fine, thanks :) I'll go through all the other things another time tomorrow

sdroege:
Seems good overall, but I'm sure there are some hidden issues left. It's just so much code :) Let's fix those when they appear

We merge this after https://github.com/gtk-rs/gir/pull/509 so that the autogenerated code is also ok?

EPashkin:
This PR still long to go.
I still don't do anything with writers asyncs (`FnOnce`, buffers passtrought, tests).
`Converter` need `AsRef` and `AsMut`, etc.

Theoretically all async need manual fix to use `Priority`, but as it can't be source of error, I currently prefer skip this for functions with `FnOnce(Result<(), Error>)` callback.

sdroege:
You could detect priority in gir and rewrite it there for async functions

EPashkin:
As `Priority` need `.to_glib()` conversion IMHO this will difficult.

sdroege:
ToGlib is no problem, FromGlib would be

sdroege:
I was also thinking about the "read problem", where nothing is preventing the implementation to modify parts outside the actually read part. The same would also apply to the Rust Read trait for example. So maybe it's save to use any AsMut there and just return the number of read bytes, or rather it's as safe as the Read trait already is.

EPashkin:
For `MemoryOutputStream` IMHO better hide all except `new_resizable`, `get_data_size`, `steal_as_bytes` and `connect_property_data_size_notify`.
@GuillaumeGomez, @sdroege Any objections?

GuillaumeGomez:
None from me.

sdroege:
Makes sense, yes

EPashkin:
Cleared `MemoryOutputStream`
On starting processing `OutputStream` found that `write_async` also need buffer pass-through.
@sdroege Please, check it.

sdroege:
👍

src/buffered_input_stream.rs
```diff
@@ -0,0 +1,50 @@
...
+use gobject_ffi;
+use ffi;
+use glib::object::IsA;
+use glib::translate::*;
+use BufferedInputStream;
+use Cancellable;
+use Error;
+
+pub trait BufferedInputStreamExtManual {
+    fn fill_async<'a, P: Into<Option<&'a Cancellable>>, Q: Fn(Result<isize, Error>) + Send + Sync + 'static>(&self, count: isize, io_priority: i32, cancellable: P, callback: Q);
```
sdroege: This is still using the old closures

src/file_i_o_stream.rs
```diff
@@ -0,0 +1,71 @@
...
+        get_type => || ffi::g_file_io_stream_get_type(),
+    }
+}
+
+pub trait FileIOStreamExt {
+    fn get_etag(&self) -> Option<String>;
+
+    fn query_info<'a, P: Into<Option<&'a Cancellable>>>(&self, attributes: &str, cancellable: P) -> Result<FileInfo, Error>;
+
+    fn query_info_async<'a, P: Into<Option<&'a Cancellable>>, Q: Fn(Result<FileInfo, Error>) + Send + Sync + 'static>(&self, attributes: &str, io_priority: i32, cancellable: P, callback: Q);
```
sdroege: And this

src/file_input_stream.rs
```diff
@@ -0,0 +1,63 @@
...
+
+    match fn {
+        get_type => || ffi::g_file_input_stream_get_type(),
+    }
+}
+
+pub trait FileInputStreamExt {
+    fn query_info<'a, P: Into<Option<&'a Cancellable>>>(&self, attributes: &str, cancellable: P) -> Result<FileInfo, Error>;
+
+    fn query_info_async<'a, P: Into<Option<&'a Cancellable>>, Q: Fn(Result<FileInfo, Error>) + Send + Sync + 'static>(&self, attributes: &str, io_priority: i32, cancellable: P, callback: Q);
```
sdroege: And this

src/file_output_stream.rs
```diff
@@ -0,0 +1,71 @@
...
+        get_type => || ffi::g_file_output_stream_get_type(),
+    }
+}
+
+pub trait FileOutputStreamExt {
+    fn get_etag(&self) -> Option<String>;
+
+    fn query_info<'a, P: Into<Option<&'a Cancellable>>>(&self, attributes: &str, cancellable: P) -> Result<FileInfo, Error>;
+
+    fn query_info_async<'a, P: Into<Option<&'a Cancellable>>, Q: Fn(Result<FileInfo, Error>) + Send + Sync + 'static>(&self, attributes: &str, io_priority: i32, cancellable: P, callback: Q);
```
sdroege: And this

sdroege:
@EPashkin Btw, did you see my comment here yesterday https://github.com/gtk-rs/gio/pull/69#issuecomment-351369144 ?

EPashkin:
Missed it. I will try to use `AsMut` in `InputStream` too.
So currently only fully processed memory streams, all others need second pass.

EPashkin:
@sdroege  We forgotten that `AsMut` impossible for reads until we use `.set_len`.
Change tactic and return buffer and count filled? Or leave as is.

EPashkin:
I meant, that we need use initialized buffer and pass `len` instead `capacity`?

sdroege:
That was my point, you can actually use `AsMut<[u8]>` here and return the number of read bytes. It's basically the same as the `Read` trait. If that is considered safe, we can consider it safe here too.

EPashkin:
Ok, I do this tomorrow, too late for today :(

sdroege:
Sure, no problem :)

EPashkin:
Now `src/input_stream.rs` and `src/output_stream.rs` is done.
Others streams still not fully.

sdroege:
:+1:

EPashkin:
Seems [Simplify test_util::run_async](https://github.com/EPashkin/gio/commit/5da2f2457571e3a3715eb350fc731350798eb4e1) was bad idea

src/file.rs
```diff
@@ -13,246 +13,225 @@ use FileQueryInfoFlags;
...
+use glib::Priority;
 use glib_ffi;
 use gobject_ffi;
-use std::mem::transmute;
 use std::ptr;
 use File;
 
 pub trait FileExtManual {
-    fn append_to_async<'a, P: Into<Option<&'a Cancellable>>, Q: Fn(Result<FileOutputStream, Error>) + Send + Sync + 'static>(&self, flags: FileCreateFlags, io_priority: i32, cancellable: P, callback: Q);
+    fn append_to_async<'a, P: Into<Option<&'a Cancellable>>, Q: FnOnce(Result<FileOutputStream, Error>) + Send + 'static>(&self, flags: FileCreateFlags, io_priority: Priority, cancellable: P, callback: Q);
```
sdroege: Why are *all* these manual? Only because of the priority? I'd rather add support to gir for this instead, this is a huge amount of manual code otherwise :)

EPashkin:
It because generated have `Result<(),Error>` instead needed

sdroege:
> It because generated have Result<(),Error> instead needed

Oh why? `g_file_append_to_finish()` returns a `FileOutputStream`

src/file.rs
EPashkin: I think that it part of https://github.com/gtk-rs/gir/issues/504

sdroege: Not really, but it's a missing feature in gir indeed. It's a normal GIO-style async function not handled by gir yet

EPashkin:
@GuillaumeGomez , @sdroege 
PR is ready for final review.

Gir.toml
```diff
@@ -43,7 +41,6 @@ generate = [
     "Gio.SettingsBindFlags",
     "Gio.SettingsSchema",
     "Gio.SimpleActionGroup",
-    "Gio.SimpleIOStream",
```
sdroege: Why do you drop this one? It's not very important for bindings in any case

src/auto/filter_input_stream.rs
```diff
@@ -31,8 +31,6 @@ pub trait FilterInputStreamExt {
 
     fn set_close_base_stream(&self, close_base: bool);
 
-    fn connect_property_base_stream_notify<F: Fn(&Self) + 'static>(&self, f: F) -> SignalHandlerId;
```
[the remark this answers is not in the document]
sdroege: I.e. generate = ["getter", "setter", "notify", "emit"]

sdroege:
Seems good for me, I'm just worried by the huge amount of manual code and the inconsistencies around the `io_priority` stuff. Maybe let's get it in, and then make sure that gir gets fixed for handling most of these cases (i.e. all but the read/write functions)

Gir.toml
EPashkin: It just moved as object to hide properties

src/auto/filter_input_stream.rs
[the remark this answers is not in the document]
EPashkin: Good idea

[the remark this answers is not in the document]
EPashkin: Added issue https://github.com/gtk-rs/gir/issues/513

EPashkin:
@GuillaumeGomez please, look at this PR too.

GuillaumeGomez:
Seems good to me as well, thanks!

PR #79, opened 2018-02-08
Fix-up Socket::create_source() implementation

sdroege:
We need to provide a callback, etc. for this function to be useful.


EPashkin:
@sdroege Looks good.
You seems forgot to add Gir.toml changes.

sdroege:
Updated

EPashkin:
@sdroege Thanks
👍

PR #83, opened 2018-02-17
More tls bindings

MathieuDuponchelle:
I only added the generated bindings, will happily fix anything that review points out, not yet comfortable enough to tell what's unsafely wrapped I'm afraid :)


src/auto/tls_interaction.rs
```diff
@@ -0,0 +1,127 @@
...
+        let cancellable = cancellable.into();
+        let cancellable = cancellable.to_glib_none();
+        unsafe {
+            let mut error = ptr::null_mut();
+            let ret = ffi::g_tls_interaction_ask_password(self.to_glib_none().0, password.to_glib_none().0, cancellable.0, &mut error);
+            if error.is_null() { Ok(from_glib(ret)) } else { Err(from_glib_full(error)) }
+        }
+    }
+
+    fn ask_password_async<'a, P: Into<Option<&'a Cancellable>>, Q: FnOnce(Result<(), Error>) + Send + 'static>(&self, password: &TlsPassword, cancellable: P, callback: Q) {
```
EPashkin: This must have `FnOnce(Result<TlsInteractionResult, Error>) ...` see `ask_password` return type,
so or ignore or make manual.
There will be also same error in others async functions for sure.

EPashkin:
@MathieuDuponchelle please rebase on new master after #84 merged

src/auto/tls_database.rs
```diff
@@ -0,0 +1,166 @@
...
+        get_type => || ffi::g_tls_database_get_type(),
+    }
+}
+
+pub trait TlsDatabaseExt {
+    fn create_certificate_handle(&self, certificate: &TlsCertificate) -> Option<String>;
+
+    fn lookup_certificate_for_handle<'a, 'b, P: Into<Option<&'a TlsInteraction>>, Q: Into<Option<&'b Cancellable>>>(&self, handle: &str, interaction: P, flags: TlsDatabaseLookupFlags, cancellable: Q) -> Result<Option<TlsCertificate>, Error>;
+
+    fn lookup_certificate_for_handle_async<'a, 'b, P: Into<Option<&'a TlsInteraction>>, Q: Into<Option<&'b Cancellable>>, R: FnOnce(Result<(), Error>) + Send + 'static>(&self, handle: &str, interaction: P, flags: TlsDatabaseLookupFlags, cancellable: Q, callback: R);
```
sdroege: These async functions are all wrong because they ignore the non-bool return value of the finish function (it should be put as first the component of the Result in the closure). Need to be written manually for now, see https://github.com/gtk-rs/gir/issues/535

src/auto/tls_database.rs
```diff
@@ -0,0 +1,166 @@
...
+
+    fn lookup_certificate_issuer<'a, 'b, P: Into<Option<&'a TlsInteraction>>, Q: Into<Option<&'b Cancellable>>>(&self, certificate: &TlsCertificate, interaction: P, flags: TlsDatabaseLookupFlags, cancellable: Q) -> Result<TlsCertificate, Error>;
+
+    fn lookup_certificate_issuer_async<'a, 'b, P: Into<Option<&'a TlsInteraction>>, Q: Into<Option<&'b Cancellable>>, R: FnOnce(Result<(), Error>) + Send + 'static>(&self, certificate: &TlsCertificate, interaction: P, flags: TlsDatabaseLookupFlags, cancellable: Q, callback: R);
+
+    //fn lookup_certificates_issued_by<'a, 'b, P: Into<Option<&'a TlsInteraction>>, Q: Into<Option<&'b Cancellable>>>(&self, issuer_raw_dn: /*Ignored*/&glib::ByteArray, interaction: P, flags: TlsDatabaseLookupFlags, cancellable: Q) -> Result<Vec<TlsCertificate>, Error>;
```
sdroege: Hm why is this not using a GBytes? Should probably add bindings for GByteArray at some point :)

src/auto/tls_interaction.rs
```diff
@@ -0,0 +1,127 @@
...
+    fn invoke_ask_password<'a, P: Into<Option<&'a Cancellable>>>(&self, password: &TlsPassword, cancellable: P) -> Result<TlsInteractionResult, Error>;
+
+    #[cfg(any(feature = "v2_40", feature = "dox"))]
+    fn invoke_request_certificate<'a, P: IsA<TlsConnection>, Q: Into<Option<&'a Cancellable>>>(&self, connection: &P, flags: TlsCertificateRequestFlags, cancellable: Q) -> Result<TlsInteractionResult, Error>;
+
+    #[cfg(any(feature = "v2_40", feature = "dox"))]
+    fn request_certificate<'a, P: IsA<TlsConnection>, Q: Into<Option<&'a Cancellable>>>(&self, connection: &P, flags: TlsCertificateRequestFlags, cancellable: Q) -> Result<TlsInteractionResult, Error>;
+
+    #[cfg(any(feature = "v2_40", feature = "dox"))]
+    fn request_certificate_async<'a, P: IsA<TlsConnection>, Q: Into<Option<&'a Cancellable>>, R: FnOnce(Result<(), Error>) + Send + 'static>(&self, connection: &P, flags: TlsCertificateRequestFlags, cancellable: Q, callback: R);
```
sdroege: Same problem with the async function, probably various others elsewhere too

src/auto/tls_connection.rs
```diff
@@ -0,0 +1,377 @@
...
+use glib::translate::*;
+use glib_ffi;
+use gobject_ffi;
+use std::boxed::Box as Box_;
+use std::mem;
+use std::mem::transmute;
+use std::ptr;
+
+glib_wrapper! {
+    pub struct TlsConnection(Object<ffi::GTlsConnection, ffi::GTlsConnectionClass>): IOStream;
```
sdroege: While we're at it, this is useless without GTlsClientConnection and/or GTlsServerConnection. Both of which look trivially autogenerateable :)

src/auto/tls_database.rs
MathieuDuponchelle: Congrats for sending me down a new rabbit hole :P

sdroege: You found the end of it already :)

MathieuDuponchelle:
Proposed #85 , will update this PR after we have merged it

src/auto/tls_interaction.rs
MathieuDuponchelle: Done

src/auto/tls_database.rs
MathieuDuponchelle: Yep, should be all good now :)

src/auto/tls_database.rs
MathieuDuponchelle: I guess :)

src/auto/tls_interaction.rs
MathieuDuponchelle: Done

src/auto/tls_connection.rs
MathieuDuponchelle: Done

src/auto/tls_client_connection.rs
```diff
@@ -0,0 +1,174 @@
...
+            if error.is_null() { Ok(from_glib_full(ret)) } else { Err(from_glib_full(error)) }
+        }
+    }
+}
+
+pub trait TlsClientConnectionExt {
+    #[cfg(any(feature = "v2_46", feature = "dox"))]
+    fn copy_session_state<P: IsA<TlsClientConnection>>(&self, source: &P);
+
+    //fn get_accepted_cas(&self) -> /*Ignored*/Vec<glib::ByteArray>;
```
sdroege: Another byte array instead of GBytes, how weird

sdroege:
Looks all good to me, but the byte array stuff annoys me. I don't think we can safely wrap that in Rust because it's refcounted and any other user of it could cause reallocation of the memory area that we might borrow currently... this should really be GBytes in the C API.

GuillaumeGomez:
Then let's go for it! Thanks everyone!

sdroege:
Something went wrong here: https://travis-ci.org/sdroege/gstreamer-rs/jobs/345089461

TlsAuthenticationMode is not generated anymore, and basically all the TLS types are missing from the Gir.toml now except for two.

sdroege:
@MathieuDuponchelle

EPashkin:
Also files like `tls_server_connection.rs` still included but not added to `auto/mod.rs`

EPashkin:
List of not generated files:
```
src/auto/tls_client_connection.rs
src/auto/tls_connection.rs
src/auto/tls_database.rs
src/auto/tls_interaction.rs
src/auto/tls_password.rs
src/auto/tls_server_connection.rs
```

sdroege:
Yes, the Gir.toml changes disappeared from the PR at some point

MathieuDuponchelle:
How weird, I have 4 files neatly unstaged here, not sure why!

PR #85, opened 2018-02-20
Async callbacks

MathieuDuponchelle:
Split in two commits, one regenerating, the other removing manual implementations of many async functions.


MathieuDuponchelle:
Hm, should have run the tests :) I'm not sure what to do here, is there a way to specify a "transform" type for parameters (Here transform int to glib::Priority)? That doesn't seem likely, otherwise I can revert to manually written bindings for these four functions I suppose

sdroege:
Those have to be manual unfortunately, unless you want to handle that somehow in gir too :)

MathieuDuponchelle:
> Those have to be manual unfortunately, unless you want to handle that somehow in gir too :)

Not sure how I would do that tbh, gir doesn't know about glib::Priority, and the function expects an int, guess I'll just have to revert (would have been nice if these constants were actually an enum)

EPashkin:
Don't worry about tests, I hope adding PR today that fix it

sdroege:
> Not sure how I would do that tbh, gir doesn't know about glib::Priority, and the function expects an int

You could specially handle arguments in async functions that are called io_priority and are of type int. Ugly hack :)

MathieuDuponchelle:
@EPashkin , curious how you will address this specific issue though ?

MathieuDuponchelle:
Ah I see, with a dirty hack, worksforme :) I'll regen once this is merged then!

EPashkin:
@MathieuDuponchelle Gir updated. Don't forget update submodule 😉

EPashkin:
@MathieuDuponchelle Thanks, but you forgot update gir submodule. (Something like `git pull origin master` in `gir` subfolder and add it to commit)
Also seems "Gio.FileIOStream", "Gio.FileInputStream", "Gio.FileOutputStream" can be generated too.

EPashkin:
As `InputStreamExt::read_bytes_async` now generated, warning appears
```
warning: unused import: `glib`
  --> src/input_stream.rs:10:5
   |
10 | use glib;
   |     ^^^^
   |
   = note: #[warn(unused_imports)] on by default
```

MathieuDuponchelle:
> Also seems "Gio.FileIOStream", "Gio.FileInputStream", "Gio.FileOutputStream" can be generated too.

Let's do that separately no?

EPashkin:
I do not insist

MathieuDuponchelle:
Should we get that in then? :)

EPashkin:
@MathieuDuponchelle LGFM, Thanks

GuillaumeGomez:
Let's go! (and thanks :wink:)

sdroege:
>>  Also seems "Gio.FileIOStream", "Gio.FileInputStream", "Gio.FileOutputStream" can be generated too.

> Let's do that separately no?

Is one of you going to do it? Otherwise please someone open a ticket so we don't forget :)

PR #86, opened 2018-02-22
Generate File{Input,Output,IO}Stream automatically

sdroege:

PR #87, opened 2018-02-23
More Tls bindings: missing files

MathieuDuponchelle:
These went missing from https://github.com/gtk-rs/gio/pull/83


EPashkin:
@MathieuDuponchelle Please, update `gir-files` submodule: it was mistakenly reverted back in #86.
Generated file must have `https://github.com/gtk-rs/gir-files @ 77d1f70` in header.

MathieuDuponchelle:
I'm a bit confused, is the new commit what you expect ?

EPashkin:
Thanks, you done it as I think.
Lately I began to think that separate commits for submodule update is not necessary if main changes contains in single commit.

MathieuDuponchelle:
> Lately I began to think that separate commits for submodule update is not necessary if main changes contains in single commit.

Would really be nice if update didn't change file headers IMO

EPashkin:
@GuillaumeGomez I propose to merge this, we have my commit to check rustdoc fix

EPashkin:
About version, maybe we need change versioning to adding single file with it? But this way we be not sure with which version file generated actually, specially in PR's
So currently I prefer current versioning.
@GuillaumeGomez, @sdroege Maybe you have better idea?

GuillaumeGomez:
I must admit that I don't have an opinion in here. Otherwise, the PR seems good to me as well so I'll wait for the debate to end before merging it. ;)

MathieuDuponchelle:
> I must admit that I don't have an opinion in here. Otherwise, the PR seems good to me as well so I'll wait for the debate to end before merging it. ;)

You shouldn't wait :) This shouldn't be blocked by a debate, by the way can i find you folks on an IRC channel? I'm on gtk-rs in on mozilla's servers, and rust on gimpnet, but if you're also there I don't recognize user names :)

GuillaumeGomez:
I'm imperio there (on both IRC channels). And since you seem to be french, I'm also on `#rust-fr` (on mozilla server).

So since @EPashkin gave the go, I'll merge. Thanks!

MathieuDuponchelle:
> I'm imperio there

That was not my first guess :)

GuillaumeGomez:
I would assume so. Someone took the pseudo on github so I went for this one. :')

PR #91, opened 2018-03-03
bind IOErrorEnum

ystreet:


sdroege:
Are some variants of the enum only available on some platforms? I can't remember

Otherwise looks good to me

GuillaumeGomez:
Strange travis error. I restarted the build.

ystreet:
@sdroege There's no `#if` guards around any of the values in the gio C headers so I think the values are available everywhere however they may not exactly be used everywhere but that's not our problem :)

EPashkin:
LGFM

GuillaumeGomez:
Thanks!

PR #98, opened 2018-04-01
Add unix socket address

antoyo:
This adds the `UnixSocketAddress` type.
I'm not sure about how `Anonymous` works, so I'm not sure how to test this variant but the other works well.


Gir.toml
```diff
@@ -83,6 +83,8 @@ generate = [
     "Gio.TlsPasswordFlags",
     "Gio.TlsRehandshakeMode",
     "Gio.TlsServerConnection",
+    "Gio.UnixSocketAddress",
+    "Gio.UnixSocketAddressType",
```
EPashkin: Both need `cfg_condition = "unix"`

src/auto/versions.txt
```diff
@@ -1,4 +1,4 @@
 All files in this directory
-were generated by gir (https://github.com/gtk-rs/gir @ d978dab)
-from gir-files (https://github.com/gtk-rs/gir-files @ 77d1f70)
+were generated by gir (https://github.com/gtk-rs/gir @ 7f5a2b5)
+from gir-files (https://github.com/gtk-rs/gir-files @ 469db10)
```
EPashkin: Change in `versions.txt` but no submodule update? That why first Travis build failed.

src/unix_socket_address.rs
```diff
@@ -0,0 +1,50 @@
+use std::ptr;
```
EPashkin: Please add header with copyrights

EPashkin:
@antoyo  Looks good for me, thanks.

src/auto/unix_socket_address.rs
```diff
@@ -0,0 +1,176 @@
...
+        }
+    }
+}
+
+pub trait UnixSocketAddressExt {
+    fn get_address_type(&self) -> UnixSocketAddressType;
+
+    fn get_is_abstract(&self) -> bool;
+
+    fn get_path(&self) -> Option<String>;
```
sdroege: Is this always a proper string, i.e. utf8? Or can it also be a filesystem path for example? I think it can

src/unix_socket_address.rs
```diff
@@ -0,0 +1,50 @@
...
+
+use ffi;
+use SocketAddress;
+use UnixSocketAddress;
+use UnixSocketAddressType;
+
+use self::AddressType::*;
+
+pub enum AddressType<'a> {
+    Path(&'a str),
```
sdroege: Shouldn't this take a `Path`?

src/unix_socket_address.rs
```diff
@@ -0,0 +1,50 @@
...
+    Anonymous,
+    Abstract(&'a str),
+    AbstractPadded(&'a str),
```
sdroege: And these are always proper `&str` or should they be `Path` or `&[u8]`?

src/unix_socket_address.rs
```diff
@@ -0,0 +1,50 @@
...
+            Path(_) => UnixSocketAddressType::Path,
+            Anonymous => UnixSocketAddressType::Anonymous,
+            Abstract(_) => UnixSocketAddressType::Abstract,
+            AbstractPadded(_) => UnixSocketAddressType::AbstractPadded,
+        }
+    }
+}
+
+pub trait UnixSocketAddressExtManual {
+    fn new_with_type(address_type: AddressType) -> UnixSocketAddress;
```
sdroege: This is a constructor, not a method. So should be simply in a `impl UnixSocketAddress` block, no new trait needed

src/auto/unix_socket_address.rs
antoyo: According to the [documentation](https://developer.gnome.org/gio/stable/GUnixSocketAddress.html#g-unix-socket-address-get-path) it can be both.

sdroege: That's inconvenient. Should return some enum then I guess, based on the type of the address?

antoyo: So, should it be a manual implementation?

antoyo:
Thanks for your comments.
I've fixed most of the issues: the one for `get_path()` should be fixed soon.

src/auto/unix_socket_address.rs
sdroege: I guess so, returning the same enum you pass to the constructor probably.

src/auto/unix_socket_address.rs
```diff
@@ -0,0 +1,151 @@
...
+glib_wrapper! {
+    pub struct UnixSocketAddress(Object<ffi::GUnixSocketAddress, ffi::GUnixSocketAddressClass>): SocketAddress, SocketConnectable;
+
+    match fn {
+        get_type => || ffi::g_unix_socket_address_get_type(),
+    }
+}
+
+impl UnixSocketAddress {
+    pub fn new(path: &str) -> UnixSocketAddress {
```
sdroege: Is this necessarily a string? AFAIU this is always a `Path`

src/auto/unix_socket_address.rs
```diff
@@ -0,0 +1,151 @@
...
+        unsafe {
+            SocketAddress::from_glib_full(ffi::g_unix_socket_address_new(path.to_glib_none().0)).downcast_unchecked()
+        }
+    }
+
+    //pub fn new_abstract(path: /*Unimplemented*/&CArray TypeId { ns_id: 0, id: 10 }) -> UnixSocketAddress {
```
sdroege: Any idea why this one is not created? It's deprecated since 2.26 from what I can see, but that would always take a `&str` AFAIU

antoyo: Probably because it has a `path_len` parameter.

src/auto/unix_socket_address.rs
antoyo: Looks like it.

antoyo: Is it okay if I do a manual implementation for now?

GuillaumeGomez: Totally.

antoyo:
So, I've fixed all the mentioned issues.
Is something missing?

sdroege:
Nothing from me

PR #102, opened 2018-04-08
Make Cancellable Send+Sync and don't generate trait for it

sdroege:
It can't be subclassed.

### DIFF
--- a/Gir.toml
+++ b/Gir.toml
@@ -16,7 +16,6 @@ generate = [
     "Gio.ApplicationFlags",
     "Gio.AskPasswordFlags",
     "Gio.BufferedOutputStream",
-    "Gio.Cancellable",
     "Gio.DataOutputStream",
     "Gio.DataStreamByteOrder",
     "Gio.DataStreamNewlineType",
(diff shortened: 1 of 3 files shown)


src/auto/versions.txt
```diff
@@ -1,4 +1,4 @@
 All files in this directory
 were generated by gir (https://github.com/gtk-rs/gir @ 6855214)
-from gir-files (https://github.com/gtk-rs/gir-files @ cf55bdc)
+from gir-files (https://github.com/gtk-rs/gir-files @ db49619)
```
EPashkin: db49619 is too old

EPashkin:
👍 without `version.txt` change

src/auto/versions.txt
sdroege: Yes, fixed

sdroege:
Without this, `Cancellable` is basically useless as you would use it to cancel something that blocks your current thread usually :)

sdroege:
Should be good to go then

tmiasko:
If you don't need `reset` I would drop it from bindings.

sdroege:
Ugly indeed :) I send another PR for that later

GuillaumeGomez:
Code looks good to me. Should I wait for the other PR before merging this one?

sdroege:
They're independent things

PR #103, opened 2018-04-08
Generate ApplicationCommandLine.

tmiasko:


sdroege:
:+1:

EPashkin:
👍 
Maybe needed manual print, print_err implementations with signle `&str` parameter,
but then we need check it for '%'

Issue #93, opened 2018-03-04
Async tests stalls


EPashkin:
Sometimes gio test stalls
https://travis-ci.org/gtk-rs/gio/builds/348980232

Tried to add delay in spawned thread but it don't helps.

```diff
diff --git a/src/test_util.rs b/src/test_util.rs
index 660e6c0..40a7380 100644
--- a/src/test_util.rs
+++ b/src/test_util.rs
@@ -4,6 +4,7 @@
 
 
 use std::sync::mpsc::{channel, Sender};
+use std::time::Duration;
 use std::thread;
 use glib::*;
 
@@ -16,6 +17,7 @@ pub fn run_async<T: Send + 'static, Q: FnOnce(Sender<T>, MainLoop) + Send + 'sta
     let (tx, rx) = channel();
 
     thread::spawn(move || {
+        thread::sleep(Duration::from_millis(100));
         c.invoke(move || {
             start(tx, l_clone);
         });
```

sdroege:
Do you have a backtrace of all threads when that happens?

EPashkin:
It seems don't happened on my computer, only seen in travis.

EPashkin:
Also not sure how get backtrace on stalling program.

sdroege:
You could attach to it with `gdb --pid=12345`. But if this is only on travis then that's not going to work

antoyo:
By the way, it's possible to run a travis build in debug mode and get a "ssh" access afterwards.
That might help to debug this issue.

sdroege:
Fixed a long time ago

PR #107, opened 2018-04-17
Avoid main context interference when running async tests.

tmiasko:
* Create separate context for each invocation instead of using global one.
* Configure thread default context before invoking closure so that
  operations will be executed in correct context.
* Avoid creating additional threads, they are unnecessary.

Fixes issue #93

### DIFF
--- a/src/test_util.rs
+++ b/src/test_util.rs
@@ -4,24 +4,23 @@
 
 
 use std::sync::mpsc::{channel, Sender};
-use std::thread;
 use glib::*;
 
 #[allow(dead_code)]
 pub fn run_async<T: Send + 'static, Q: FnOnce(Sender<T>, MainLoop) + Send + 'static>(start: Q) -> T {
-    let l = MainLoop::new(None, false);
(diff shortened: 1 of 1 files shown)


sdroege:
:+1:

src/test_util.rs
```diff
@@ -4,25 +4,24 @@
...
 
     let (tx, rx) = channel();
 
-    thread::spawn(move || {
-        c.invoke(move || {
-            start(tx, l_clone);
-        });
+    c.invoke(move || {
+        let c = l_clone.get_context().unwrap();
+        c.push_thread_default();
```
sdroege: I would do that simply at the top/bottom of the run_async function to have it for the whole runtime of the test

tmiasko: Updated accordingly.

sdroege:
@GuillaumeGomez Let's get this in then, together with Phil's PR this should solve most of the failures

GuillaumeGomez:
Agreed. Just waiting for CIs. :)

PR #108, opened 2018-04-18
Add futures variants for all async functions.

sdroege:
This depends on https://github.com/gtk-rs/gir/pull/582 and https://github.com/gtk-rs/glib/pull/314

For an example usage see the memory input stream unit test, I'll also add a more involved example to the examples repository once everything is merged


EPashkin:
@GuillaumeGomez Please restart CI here

sdroege:
This currently uses my own branch of glib, switching back :)

sdroege:
I'm also currently working on rebasing this and making it work with @tmiasko's changes (except for the rename of all the files)

sdroege:
new cleaned up version of everything, this handles global functions too now

sdroege:
This should be good to go now

src/gio_future.rs
```diff
@@ -0,0 +1,107 @@
...
+// Licensed under the MIT license, see the LICENSE file or <http://opensource.org/licenses/MIT>
+
+use futures_channel::oneshot;
+use futures_core::task::Context;
+use futures_core::{Async, Future};
+
+use glib;
+use Cancellable;
+
+/// Represents a GIO asynchronous operation as a future
```
EPashkin: This comment and for new better removed and added to lgpl-docs later

sdroege: Why?

EPashkin: Because travis don't allow class and function comment in gio at all.

sdroege: Can't we just disable that? We might want to add more docs in the future, and this is not really part of the gir-extracted docs.

EPashkin: We have https://raw.githubusercontent.com/gtk-rs/lgpl-docs/master/gio/docs.md specially for manual comments.

sdroege:
Ah I did not know that. I'll remove the docs here then

src/gio_future.rs
sdroege: Done

src/lib.rs
```diff
@@ -31,6 +41,9 @@ mod socket_listener;
 #[cfg(any(unix, feature = "dox"))]
 mod unix_socket_address;
 mod file;
+mod pollable_input_stream;
+mod pollable_output_stream;
+mod file;
```
EPashkin: Seems this effect of many rebases: second `mod file`

src/lib.rs
```diff
@@ -57,9 +70,17 @@ pub mod prelude {
     #[cfg(any(unix, feature = "dox"))]
     pub use unix_socket_address::{UnixSocketAddressPath, UnixSocketAddressExtManual};
     pub use file::FileExtManual;
+    pub use pollable_input_stream::PollableInputStreamExtManual;
+    pub use pollable_output_stream::PollableOutputStreamExtManual;
+    pub use file::FileExtManual;
```
EPashkin: Same here

src/lib.rs
EPashkin: Seem, commit `Implement File::replace_contents_async manually` was not fully removed from this PR.

sdroege: Fixed both, thanks

sdroege:
Rebased to remove all the other PRs that are in master now. This should be good to go.

And as a next step we can regenerate all crate with latest GIR (+ file renames).

EPashkin:
It contains new generated code but no submodule updates, so travis fails.
I restarted appveyor, it IMHO be fixed now.

sdroege:
> It contains new generated code but no submodule updates, so travis fails.

I see, but we can't use latest GIR because that already contains all the file renames :) My suggestion would be to get this here in first, and then we fix up all these things by regenerating everything one by one

EPashkin:
Don't you already do regen with latest gir?
I have only `versions.txt` change on local update.

EPashkin:
Hm, there both old and new files here.

sdroege:
No, because the latest GIR would pull in the file renames and then we have a *huge* diff with not only the futures addition but also the renames and it would be very hard to follow from the GIT history what exactly happened/changed.

sdroege:
Hrm indeed. We have everything at once here already, I must've updated the GIR version at some point.

Ok so how to proceed? First without renames, then another with renames? Or all at once?

EPashkin:
https://github.com/gtk-rs/gio/pull/108/commits/8f4831276ad551fc92d809cd8cb64358dd28cc14 contains both `file_i_o_stream.rs`+`file_io_stream.rs`, `i_o_stream.rs`+`io_stream.rs`, `simple_i_o_stream.rs`+`simple_io_stream.rs`

EPashkin:
IMHO better remove old version in same commit, so git show that file moved.

tmiasko:
I would fix the commit https://github.com/gtk-rs/gio/commit/8f4831276ad551fc92d809cd8cb64358dd28cc14, then in additional commit rm -r auto, update submodules and regenerate with renames so that CI pass.

sdroege:
Ok let's do that then

sdroege:
This should be good now. One commit for the futures addition, another for the renames and updating to latest GIR.

sdroege:
Yes let's not enable more tests until all tests are actually succeeding on CI again

sdroege:
travis is green, nice

EPashkin:
Apveyor good too.

GuillaumeGomez:
Good for me as well. Thanks!

sdroege:
Thanks! I send PRs for everything else this afternoon

PR #109, opened 2018-04-19
Implement File::replace_contents_async manually

sdroege:
The contents need to be owned until the callback has triggered.


sdroege:
Extracted from https://github.com/gtk-rs/gio/pull/108 as it's unrelated

EPashkin:
👍 Even if appveyor fail

GuillaumeGomez:
Thanks!

PR #110, opened 2018-04-21
Generate D-Bus types.

tmiasko:
Dependencies:

* [x] GString support in glib crate - https://github.com/gtk-rs/glib/pull/317.
* [x] Turning VariantTy into owned type in generator (occurs only in one function so could be ignored as well).


GuillaumeGomez:
Wo, quite the old PR! Not sure if still of actuality but just in case:

cc @sdroege 

Don't hesitate to ping someone if you see no activity!

sdroege:
Needs to be rebased and regenerated with latest GIR I guess

sdroege:
Looks generally good (I only shortly looked over it, I don't have the time to go through each function in detail now) but is of course very low-level binding and not very convenient to use. It would be good to have some example/test code for this, e.g. in `gtk-examples`.

PR #111, opened 2018-04-21
Remove unneded futures-executor dependency

sdroege:


EPashkin:
@sdroege Thanks, LGFM

PR #123, opened 2018-06-12
Add new types

GuillaumeGomez:
cc @EPashkin @sdroege


src/desktop_app_info.rs
```diff
@@ -0,0 +1,37 @@
...
+// Licensed under the MIT license, see the LICENSE file or <http://opensource.org/licenses/MIT>
+
+use DesktopAppInfo;
+use libc;
+use ffi;
+use glib_ffi;
+use glib::translate::*;
+
+impl DesktopAppInfo {
+    pub fn search(search_string: &str) -> Vec<Vec<String>> {
```
GuillaumeGomez: A double check on this function implementation would be very useful!

EPashkin:
Seems `DesktopAppInfo` need `cfg(not(windows))` it not links for me :(

EPashkin:
also need add `#![allow(deprecated)]` to lib.rs for deprecated eject futures.

src/desktop_app_info.rs
EPashkin: Looks good for me

src/auto/zlib_compressor.rs
```diff
@@ -0,0 +1,130 @@
...
+use glib::translate::*;
+use glib_ffi;
+use gobject_ffi;
+use std::boxed::Box as Box_;
+use std::mem;
+use std::mem::transmute;
+use std::ptr;
+
+glib_wrapper! {
+    pub struct ZlibCompressor(Object<ffi::GZlibCompressor, ffi::GZlibCompressorClass>);
```
sdroege: Without `GConverter` this is useless

src/auto/zlib_decompressor.rs
```diff
@@ -0,0 +1,93 @@
...
+use glib::translate::*;
+use glib_ffi;
+use gobject_ffi;
+use std::boxed::Box as Box_;
+use std::mem;
+use std::mem::transmute;
+use std::ptr;
+
+glib_wrapper! {
+    pub struct ZlibDecompressor(Object<ffi::GZlibDecompressor, ffi::GZlibDecompressorClass>);
```
sdroege: Without `GConverter` this is useless (for new types, check in the docs which interfaces it implements and parent classes)

src/desktop_app_info.rs
sdroege: :+1:

GuillaumeGomez:
Updated.

GuillaumeGomez:
Is it ok for you @sdroege and @EPashkin?

EPashkin:
@GuillaumeGomez IMHO better or implement main Converter function manually (There will be some problem, I don't remember which)  or remove Zlib[De]compressor

GuillaumeGomez:
What problem? Seems like it got generated fine.

src/auto/converter.rs
```diff
@@ -0,0 +1,37 @@
...
+glib_wrapper! {
+    pub struct Converter(Object<ffi::GConverter, ffi::GConverterIface>);
+
+    match fn {
+        get_type => || ffi::g_converter_get_type(),
+    }
+}
+
+pub trait ConverterExt {
+    //fn convert(&self, inbuf: &[u8], outbuf: &[u8], flags: /*Ignored*/ConverterFlags) -> Result<(/*Ignored*/ConverterResult, usize, usize), Error>;
```
EPashkin: Without this function all xxConvert unusable

EPashkin: including [de]compressors

GuillaumeGomez:
Generated as well. Didn't this one when checking missing types...

src/auto/converter.rs
```diff
@@ -0,0 +1,48 @@
...
+    fn convert(&self, inbuf: &[u8], outbuf: &[u8], flags: ConverterFlags) -> Result<(ConverterResult, usize, usize), Error>;
+
+    fn reset(&self);
+}
+
+impl<O: IsA<Converter>> ConverterExt for O {
+    fn convert(&self, inbuf: &[u8], outbuf: &[u8], flags: ConverterFlags) -> Result<(ConverterResult, usize, usize), Error> {
+        let inbuf_size = inbuf.len() as usize;
+        let outbuf_size = outbuf.len() as usize;
+        unsafe {
```
EPashkin: As I remember `outbuf` need be mutable as converter write data to it,
something like `buffer` in `InputStreamExtManual::read`,
`inbuf` also better be `AsRef<[u8]>` as in `FileExtManual::replace_contents_async` but maybe it unneeded.

sdroege: Yes, at least make it a `&mut [u8]` but otherwise `AsRef<[u8]>` / `AsMut<[u8]>` would be more generic. I'd suggest using references though as this function does not take ownership of the objects, and the user can always get the slice reference themselves

sdroege:
Looks good to me apart from the `GConverter` function, fix that and get it in then :)

EPashkin:
IMHO all good, thanks

GuillaumeGomez:
Thanks for your reviews!

sdroege:
Looks good to me too

PR #124, opened 2018-06-17
New types v3

GuillaumeGomez:
cc @EPashkin @sdroege


src/auto/bytes_icon.rs
```diff
@@ -0,0 +1,79 @@
...
+            from_glib_full(ffi::g_bytes_icon_new(bytes.to_glib_none().0))
+        }
+    }
+}
+
+pub trait BytesIconExt {
+    #[cfg(any(feature = "v2_38", feature = "dox"))]
+    fn get_bytes(&self) -> Option<glib::Bytes>;
+
+    fn get_property_bytes(&self) -> Option<glib::Bytes>;
```
EPashkin: This property need version 2_38 to be hidden

src/lib.rs
```diff
@@ -1,8 +1,9 @@
-// Copyright 2013-2016, The Gtk-rs Project Developers.
+// Copyright 2013-2018, The Gtk-rs Project Developers.
 // See the COPYRIGHT file at the top-level directory of this distribution.
 // Licensed under the MIT license, see the LICENSE file or <http://opensource.org/licenses/MIT>
 
 #![allow(deprecated)]
+#![feature(extern_prelude)]
```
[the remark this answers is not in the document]
GuillaumeGomez: Apparently not. But without it, it doesn't work on nightly anymore. I'll need a new cfg, yeay, breaking changes...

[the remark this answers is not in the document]
EPashkin: Its gir error, it don't add `use glib;` to `auto\loadable_icon.rs`.

src/auto/bytes_icon.rs
GuillaumeGomez: Why? https://developer.gnome.org/gobject/stable/gobject-The-Base-Object-Type.html#g-object-get-property There is no minimum version...

EPashkin: It minimum version for class itself and we already have `https://developer.gnome.org/gio/stable/GBytesIcon.html#g-bytes-icon-get-bytes`, so better hide property by setting same version to it.
Maybe this need do gir automatically.

PS. Strange that no set property generated

EPashkin: Ah understand why no set, it Write/Construct only.

GuillaumeGomez: Ok, I add the version then.

GuillaumeGomez:
@EPashkin: Like this?

EPashkin:
Yes, all looks good. Thanks.

GuillaumeGomez:
Waiting for @sdroege now. :)

sdroege:
Looks good to me

PR #125, opened 2018-06-19
Add new types

GuillaumeGomez:
cc @sdroege @EPashkin


src/auto/subprocess.rs
```diff
@@ -0,0 +1,492 @@
...
+    pub struct Subprocess(Object<ffi::GSubprocess>);
+
+    match fn {
+        get_type => || ffi::g_subprocess_get_type(),
+    }
+}
+
+impl Subprocess {
+    //#[cfg(any(feature = "v2_40", feature = "dox"))]
+    //pub fn new<'a, P: Into<Option<&'a Error>>>(flags: SubprocessFlags, error: P, argv0: &str, : /*Unknown conversion*//*Unimplemented*/Fundamental: VarArgs) -> Subprocess {
```
GuillaumeGomez: I have a crate that we could use to give `VarArgs`: https://crates.io/crates/va_list-rs (apparently, you can just give a `va_list` as a `VarArgs` argument so why not enjoying this?). Should we give it a try?

EPashkin: IMHO it worth to try but not in this PR (or if PR not merged until next release)

sdroege: Also we discussed that a while ago on IRC and I believe there are some safety concerns there. Have to look again, I can't remember :)

GuillaumeGomez: I don't intend to put it in this PR but since I had the chance to talk about it... 😛

src/auto/inet_address_mask.rs
```diff
@@ -0,0 +1,183 @@
...
+    #[inline]
+    fn fmt(&self, f: &mut fmt::Formatter) -> fmt::Result {
+        write!(f, "{}", InetAddressMaskExt::to_string(self))
+    }
+}
+
+pub trait InetAddressMaskExt {
+    fn equal(&self, mask2: &InetAddressMask) -> bool;
+
+    fn get_address(&self) -> Option<InetAddress>;
```
sdroege: Can this really return `None`?

src/auto/subprocess.rs
sdroege: Also the `newv` variant is equivalent and more Rust-y :)

src/auto/subprocess.rs
```diff
@@ -0,0 +1,492 @@
...
+    //    unsafe { TODO: call ffi::g_subprocess_new() }
+    //}
+
+    #[cfg(any(feature = "v2_40", feature = "dox"))]
+    pub fn newv(argv: &[&std::path::Path], flags: SubprocessFlags) -> Result<Subprocess, Error> {
```
sdroege: The `argv` should be an `OsString` AFAIU

src/auto/subprocess_launcher.rs
```diff
@@ -0,0 +1,220 @@
...
+    fn set_stdout_file_path<P: AsRef<std::path::Path>>(&self, path: P);
+
+    #[cfg(any(feature = "v2_40", feature = "dox"))]
+    fn setenv<P: AsRef<std::path::Path>, Q: AsRef<std::path::Path>>(&self, variable: P, value: Q, overwrite: bool);
+
+    //#[cfg(any(feature = "v2_40", feature = "dox"))]
+    //fn spawn(&self, error: &mut Error, argv0: &str, : /*Unknown conversion*//*Unimplemented*/Fundamental: VarArgs) -> Option<Subprocess>;
+
+    #[cfg(any(feature = "v2_40", feature = "dox"))]
+    fn spawnv(&self, argv: &[&std::path::Path]) -> Result<Subprocess, Error>;
```
sdroege: `OsString` as above

src/auto/subprocess_launcher.rs
```diff
@@ -0,0 +1,220 @@
...
+    fn set_stderr_file_path<P: AsRef<std::path::Path>>(&self, path: P);
+
+    #[cfg(any(feature = "v2_40", feature = "dox"))]
+    fn set_stdin_file_path(&self, path: &str);
+
+    #[cfg(any(feature = "v2_40", feature = "dox"))]
+    fn set_stdout_file_path<P: AsRef<std::path::Path>>(&self, path: P);
+
+    #[cfg(any(feature = "v2_40", feature = "dox"))]
+    fn setenv<P: AsRef<std::path::Path>, Q: AsRef<std::path::Path>>(&self, variable: P, value: Q, overwrite: bool);
```
sdroege: `OsString` as above (or rather they should all be `&OsStr`, there's a gir parameter for this)

src/auto/subprocess_launcher.rs
```diff
@@ -0,0 +1,220 @@
...
+    fn take_stderr_fd(&self, fd: i32);
+
+    #[cfg(any(feature = "v2_40", feature = "dox"))]
+    fn take_stdin_fd(&self, fd: i32);
+
+    #[cfg(any(feature = "v2_40", feature = "dox"))]
+    fn take_stdout_fd(&self, fd: i32);
+
+    #[cfg(any(feature = "v2_40", feature = "dox"))]
+    fn unsetenv<P: AsRef<std::path::Path>>(&self, variable: P);
```
sdroege: `OsString`

sdroege:
Looks generally good except for those comments above. Thanks for working on this :) This will become a good release for GIO!

GuillaumeGomez:
I hope it'll. I need to understand what's happening on the appveyor failure as well. Some duplicate implementations apparently...

EPashkin:
Duplicate trait happened on travis too.
Only objects have buildin `Eq` (added long ago https://github.com/gtk-rs/glib/commit/10745adc744d755113ebba876d120a482f07d066).
Not sure what we can do (IMHO ignoring `equals` remove generated `Eq` but it not rigth)

GuillaumeGomez:
I'll test and we'll see. In worst case, I'll just copy/paste the whole file and ignore its generation...

sdroege:
The `Eq` thing is somewhat related to https://github.com/gtk-rs/glib/issues/247

We need a way to opt-out of the automatic `Eq`, `Debug`, etc impls and have custom ones.

sdroege:
> I'll test and we'll see. In worst case, I'll just copy/paste the whole file and ignore its generation...

You can ignore the `equal` function in gir too probably. But it's all not a solution really :)

src/auto/subprocess_launcher.rs
```diff
@@ -0,0 +1,220 @@
...
+    fn setenv<P: AsRef<std::path::Path>, Q: AsRef<std::path::Path>>(&self, variable: P, value: Q, overwrite: bool);
+
+    //#[cfg(any(feature = "v2_40", feature = "dox"))]
+    //fn spawn(&self, error: &mut Error, argv0: &str, : /*Unknown conversion*//*Unimplemented*/Fundamental: VarArgs) -> Option<Subprocess>;
+
+    #[cfg(any(feature = "v2_40", feature = "dox"))]
+    fn spawnv(&self, argv: &[&std::path::Path]) -> Result<Subprocess, Error>;
+
+    #[cfg(any(feature = "v2_40", feature = "dox"))]
+    fn take_fd(&self, source_fd: i32, target_fd: i32);
```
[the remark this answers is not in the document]
sdroege: These fd functions are also unsafe, as you can safely create a random fd (it's just a type alias for an integer) and then having GLib use that will cause problems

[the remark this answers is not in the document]
sdroege: I mean they must be marked as `unsafe`

src/auto/inet_address_mask.rs
GuillaumeGomez: I suppose it [cannot](https://github.com/GNOME/glib/blob/2dce3ce125cd0a58d4f49517dcd451857d62bdbd/gio/ginetaddressmask.c#L384).

GuillaumeGomez:
Updated.

src/subprocess_launcher.rs
```diff
@@ -0,0 +1,51 @@
+#[cfg(any(feature = "v2_40", feature = "dox"))]
+use SubprocessLauncher;
+use ffi;
+use glib;
+use glib::object::IsA;
+use GtkRawFd;
```
sdroege: I don't like this one, and it will also make the API unsafe as you can pass random integers in there now. There's a reason why I said to use the proper traits for this :)

src/subprocess_launcher.rs
```diff
@@ -0,0 +1,51 @@
...
+
+pub trait SubprocessLauncherExtManual {
+    #[cfg(any(feature = "v2_40", feature = "dox"))]
+    fn take_fd(&self, source_fd: GtkRawFd, target_fd: GtkRawFd);
```
sdroege: These should all take a https://doc.rust-lang.org/nightly/std/os/unix/io/trait.IntoRawFd.html on UNIX and https://doc.rust-lang.org/nightly/std/os/windows/io/trait.IntoRawHandle.html on Windows to be safe (the safety aspect is then moved to the implementors of that trait and people can safely pass various types in there)

sdroege: See e.g. https://github.com/sdroege/gstreamer-rs/blob/master/gstreamer/src/bus_windows.rs and https://github.com/sdroege/gstreamer-rs/blob/master/gstreamer/src/bus_unix.rs

EPashkin: or https://github.com/gtk-rs/gio/blob/master/src/socket.rs

EPashkin: Was wrong, my link only for sockets

sdroege: Basically same thing though, yes

GuillaumeGomez: Yes, I had unsafe in mind, but I don't get why I didn't think about `AsRawFd` and `AsRawHandle`...

sdroege: Into, not As :) this takes ownership. And it does not have to be unsafe then because the trait impls already take care of that being safe

GuillaumeGomez: True!

GuillaumeGomez:
Okay, the safe version is now here!

GuillaumeGomez:
After multiple windows fixes, does it seem good?

EPashkin:
Travis don't think so 😭

EPashkin:
By way, seems manual function in `SubProcessLauncher` is not_windows only, I can't find it in `libgio-2.0-0.dll`  while `g_subprocess_launcher_new` is present

GuillaumeGomez:
Another bug in `gio` I assume... I just switch to `cfg(not(windows))` for `SubProcessLauncher` then?

sdroege:
Apart from the Windows stuff and the async function ownership problem this looks good to me now

src/auto/subprocess.rs
```diff
@@ -0,0 +1,492 @@
...
+
+    #[cfg(feature = "futures")]
+    #[cfg(any(feature = "v2_40", feature = "dox"))]
+    fn communicate_async_future<'a, P: Into<Option<&'a glib::Bytes>>>(&self, stdin_buf: P) -> Box_<futures_core::Future<Item = (Self, (glib::Bytes, glib::Bytes)), Error = (Self, Error)>>;
+
+    #[cfg(any(feature = "v2_40", feature = "dox"))]
+    fn communicate_utf8<'a, 'b, P: Into<Option<&'a str>>, Q: Into<Option<&'b Cancellable>>>(&self, stdin_buf: P, cancellable: Q) -> Result<(Option<String>, Option<String>), Error>;
+
+    #[cfg(any(feature = "v2_40", feature = "dox"))]
+    fn communicate_utf8_async<'a, 'b, P: Into<Option<&'a str>>, Q: Into<Option<&'b Cancellable>>, R: FnOnce(Result<(String, String), Error>) + Send + 'static>(&self, stdin_buf: P, cancellable: Q, callback: R);
```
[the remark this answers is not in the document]
GuillaumeGomez: So I just switch its lifetime to `'b` and we're good or should I move it to `'static`?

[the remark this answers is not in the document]
sdroege: Needs to be `String` instead of `&str`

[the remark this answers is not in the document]
GuillaumeGomez: Or that haha.

EPashkin:
Futures bug now in beta too :(
Also errors and warnings in stable https://travis-ci.org/gtk-rs/gio/jobs/395928392 IMHO legit and need fix

sdroege:
> Also errors and warnings in stable https://travis-ci.org/gtk-rs/gio/jobs/395928392 IMHO legit and need fix

These are missing imports in the `subprocess.rs` in this PR (@GuillaumeGomez). And some type mismatches in there. All the other futures code seems to be fine.

Same for the beta problems.

Gir.toml
```diff
@@ -579,6 +579,12 @@ status = "generate"
         [[object.function.parameter]]
         name = "argv"
         string_type = "os_string"
+    [[object.function]]
+    name = "communicate_utf8_async"
+    ignore = true
+    [[object.function]]
+    name = "communicate_utf8_async_future"
```
sdroege: You don't have to ignore that one here. It is generated as a result of the non-futures function and does not exist as such in C

src/subprocess.rs
```diff
@@ -0,0 +1,76 @@
...
+    }
+
+    #[cfg(feature = "futures")]
+    #[cfg(any(feature = "v2_40", feature = "dox"))]
+    fn communicate_utf8_async_future<P: Into<Option<String>>>(&self, stdin_buf: P) -> Box_<futures_core::Future<Item = (Self, (String, String)), Error = (Self, Error)>> {
+        use GioFuture;
+        use send_cell::SendCell;
+
+        let stdin_buf = stdin_buf.into();
+        let stdin_buf = stdin_buf.map(ToOwned::to_owned);
```
sdroege: It already is a `String`, there's nothing to convert here. But you could make it take a `&str` and do the copy internally here like you do already

src/subprocess.rs
```diff
@@ -0,0 +1,76 @@
...
+
+#[cfg(any(feature = "v2_40", feature = "dox"))]
+impl<O: IsA<Subprocess> + IsA<glib::object::Object> + Clone + 'static> SubprocessExtManual for O {
+    #[cfg(any(feature = "v2_40", feature = "dox"))]
+    fn communicate_utf8_async<'a, P: Into<Option<String>>, Q: Into<Option<&'a Cancellable>>, R: FnOnce(Result<(String, String), Error>) + Send + 'static>(&self, stdin_buf: P, cancellable: Q, callback: R) {
+        let stdin_buf = stdin_buf.into();
+        let stdin_buf = stdin_buf.to_glib_none();
+        let cancellable = cancellable.into();
+        let cancellable = cancellable.to_glib_none();
+        let user_data: Box<Box<(R, *mut i8)>> = Box::new(Box::new((callback, stdin_buf.0)));
```
sdroege: The `stdin_buf` has to be the result of `to_glib_full()` as you want to keep it alive until the trampoline ran and then free it there.

GuillaumeGomez:
Updated!

src/subprocess.rs
```diff
@@ -0,0 +1,86 @@
+#[cfg(any(feature = "v2_40", feature = "dox"))]
+use Subprocess;
+#[cfg(any(feature = "v2_40", feature = "dox"))]
+use Cancellable;
+#[cfg(any(feature = "v2_40", feature = "dox"))]
+use Error;
+#[cfg(any(feature = "v2_40", feature = "dox"))]
+use ffi;
```
EPashkin: This better have same cfg as functions, see warning in appveyor

src/subprocess_launcher.rs
```diff
@@ -0,0 +1,62 @@
+#[cfg(any(feature = "v2_40", feature = "dox"))]
+use SubprocessLauncher;
+#[cfg(any(feature = "v2_40", feature = "dox"))]
+use ffi;
```
EPashkin: This better have same cfg as functions, see warning in appveyor

src/subprocess_launcher.rs
```diff
@@ -0,0 +1,62 @@
...
+#[cfg(any(feature = "v2_40", feature = "dox"))]
+use glib;
+#[cfg(any(feature = "v2_40", feature = "dox"))]
+use glib::object::IsA;
+#[cfg(any(all(feature = "v2_40", not(windows)), all(feature = "dox", not(windows))))]
+use std::os::unix::io::IntoRawFd;
+#[cfg(all(feature = "dox", windows))]
+pub trait IntoRawFd {}
```
EPashkin: IMHO better use `unix` and `not(unix)` instead

EPashkin:
Looks good, thanks

sdroege:
Go for it then :)

GuillaumeGomez:
I want to update following @EPashkin's comments first. Then I merge. :) Thanks a lot to both of you for your reviews!

src/subprocess_launcher.rs
```diff
@@ -0,0 +1,65 @@
...
+use glib;
+#[cfg(any(feature = "v2_40", feature = "dox"))]
+use glib::object::IsA;
+#[cfg(any(all(feature = "v2_40", not(windows)), all(feature = "dox", unix)))]
+use std::os::unix::io::IntoRawFd;
+#[cfg(all(feature = "dox", not(unix)))]
+/// Replacement for "real" [`IntoRawFd`] trait for non-unix targets.
+///
+/// [`IntoRawFd`]: https://doc.rust-lang.org/std/os/unix/io/trait.IntoRawFd.html
+pub trait IntoRawFd {}
```
EPashkin: Your check don't allow doc comments here, you need add it to lgpl-docs, but not sure that it restored in right place.
Also IMHO better not mixing `windows` and `unix` and use only `unix` (in all this file)

EPashkin:
for `futures` all v0.2.* yanked

sdroege:
> for futures all v0.2.* yanked

Great! Complained about that here, let's see what they reply https://users.rust-lang.org/t/futures-0-2-has-been-moved-to-futures-preview/18329/2?u=slomo

sdroege:
Maybe only needs moving to the futures-preview crates instead, I'll check later

src/subprocess_launcher.rs
```diff
@@ -1,36 +1,36 @@
 #[cfg(any(feature = "v2_40", feature = "dox"))]
 use SubprocessLauncher;
-#[cfg(any(all(feature = "v2_40", not(windows)), feature = "dox"))]
+#[cfg(any(all(feature = "v2_40", not(unix)), feature = "dox"))]
```
EPashkin: This is wrong change, as windows ~= not(unix)

GuillaumeGomez: Oups indeed!

GuillaumeGomez:
I fixed conditions.

sdroege:
Currently preparing gio-futures PR, then this can be merged IMHO

EPashkin:
@GuillaumeGomez Docs on windows failed
https://ci.appveyor.com/project/GuillaumeGomez/gio/build/job/ujoevof6anec5ur7
```
error: string literal with a suffix is invalid
  --> src\subprocess_launcher.rs:12:9
   |
12 | #[doc = "Replacement for "real" [`IntoRawFd`] trait for non-unix targets.
   |         ^^^^^^^^^^^^^^^^^^^^^^
error: unexpected token: `" [`IntoRawFd`] trait for non-unix targets.
[`IntoRawFd`]: https://doc.rust-lang.org/std/os/unix/io/trait.IntoRawFd.html"`
  --> src\subprocess_launcher.rs:12:31
   |
12 |   #[doc = "Replacement for "real" [`IntoRawFd`] trait for non-unix targets.
   |  _______________________________^
13 | |
14 | | [`IntoRawFd`]: https://doc.rust-lang.org/std/os/unix/io/trait.IntoRawFd.html"]
   | |_____________________________________________________________________________^ unexpected token after this
error: Could not document `gio`.

```

EPashkin:
.. and doc comments will be not accepted anyway it need be added to lgpl-docs

GuillaumeGomez:
This is really annoying. :-/ I'll open a PR to remove it from here.

PR #126, opened 2018-06-23
Remove callback guard

EPashkin:
Part of https://github.com/gtk-rs/glib/pull/250

cc @GuillaumeGomez, @sdroege


EPashkin:
Sorry, this better do after #125

PR #134, opened 2018-07-08
Add new types

GuillaumeGomez:
cc @sdroege @EPashkin


src/auto/task.rs
```diff
@@ -0,0 +1,333 @@
...
+use gobject_ffi;
+#[cfg(any(feature = "v2_44", feature = "dox"))]
+use std::boxed::Box as Box_;
+use std::mem;
+#[cfg(any(feature = "v2_44", feature = "dox"))]
+use std::mem::transmute;
+use std::ptr;
+
+glib_wrapper! {
+    pub struct Task(Object<ffi::GTask, ffi::GTaskClass>);
```
sdroege: This is not really that useful, it needs deeper integration than what GIR could possibly autogenerate. It probably also makes most sense with some kind of futures integration (because that's exactly what a `GTask` is).

PR #135, opened 2018-07-14
remove versions from git dependencies

vhdirk:
Same as https://github.com/gtk-rs/pangocairo/pull/20
I'll have to do this for quite a few projects.


GuillaumeGomez:
Thanks!

PR #139, opened 2018-07-23
Use a const pointer for the slice returned from g_unix_socket_address_get_path()

sdroege:
And never pass NULL to slice::from_raw_part(), it's undefined behaviour


GuillaumeGomez:
Thanks!

PR #141, opened 2018-08-08
Fix compilation with two-phase-borrows (aka NLL)

sdroege:


GuillaumeGomez:
And the last one. Thanks!

Issue #149, opened 2018-09-13
GInputStream / GOutputStream read_all()/write_all() bindings wrong in error case


sdroege:
In case of error, these functions return still how much was actually written as some part of the buffer might've been successfully written (especially in the `EWOULDBLOCK` case).

See docs

```
As a special exception to the normal conventions for functions that use GError, if this function returns
FALSE (and sets error ) then bytes_written will be set to the number of bytes that were successfully
written before the error was encountered. This functionality is only available from C. If you need it from
another language then you must write your own loop around g_output_stream_write().
```

This needs manual bindings for these functions, and does affect the sync and async variants.

EPashkin:
@sdroege Thanks.

PR #150, opened 2018-09-14
Fix OutputStream::write_all()

sdroege:
It can return an error *and* can have bytes written out. We have to
return the written bytes in any case, and did so already for the async
variants and InputStream::read_all().

Fixes https://github.com/gtk-rs/gio/issues/149


sdroege:
@GuillaumeGomez All green except for a network failure on one travis instance

GuillaumeGomez:
Good enough for me. Thanks! :)

Issue #160, opened 2018-10-15
Provide a way to use std::io::{Read,Write} and gio::{Input,Output}Stream interchangeably


fkrull:
When you have a `std::io::Read`, it should be possible to use it with gio-based APIs that require a `gio::InputStream`. Conversely, when all you have is a `gio::InputStream`, it should be possible to use it with APIs using `std::io::Read`. The same goes `std::io::Write` and `gio::OutputStream`.

When mixing APIs using the Rust types and the gio types, interaction between streams gets hairy. Writing a wrapper for `gio::{Input,Output}Stream` that implements Read or Write is straightforward enough, but it's also generally useful so IMO it should be included here. However, if all you have is an `std::io::Read`, the only way I've found to get that to an API that takes a `gio::InputStream` is to dump the entire stream into a temporary file and re-open that with gio. I think it should be possible to write a gobject subclass of GInputStream -- either in unsafe Rust or in C -- that simply wraps a `std::io::Read`. This class would then have to be exposed back to Rust in some form.

So in short, things I'd love to see:
* a wrapper for `gio::InputStream` that implements `std::io::Read`
* a wrapper for `gio::OutputStream` that implements `std::io::Write`
* a subclass of `GInputStream` that wraps a `std::io::Read` and can be passed to wrapped APIs
* a subclass of `GOutputStream` that wraps a `std::io::Write` and can be passed to wrapped APIs

sdroege:
For the first two, you could add an impl for `Read`/`Write` directly on `GInputStream`/`GOutputStream`. Do you want to provide a PR for that?

The other two are a bit more involved as the subclassing story in Rust is not great yet. You can either do it with a bit of unsafe code doing basically the same as one would do in C (which is probably preferred for this case here), or around https://github.com/gtk-rs/gobject-subclass for a generic solution.

sdroege:
> Well, I did say it should be easy, so I'll take a stab at the trait impls later.

Feel free to ask for help here if you get stuck somewhere :)

> As for the other part, I'm not very familiar with either gobject for unsafe Rust. I kinda want to give it a try, buuuuut we'll see.

Understandable, it's not exactly trivial. If you want to read up on things a bit, see my blog post here: https://coaxion.net/blog/2017/09/exporting-a-gobject-c-api-from-rust-code-and-using-it-from-c-python-javascript-and-others/

And as above, if you want to try and learn, feel free to ask here. I'd be happy to help.

fkrull:
[...]

PR #162, opened 2018-10-23
Add std::io::{Read,Write} wrappers

fkrull:
See #160. I've only done the Read impl for now because figuring out the API was trickier than I expected, so I wanted to get some feedback first.

So it turns out you can't just implement Read for any type that implements InputStreamExt (or some other trait). I came up with a few ideas for that:

* Implement the trait for the InputStream struct, and all the structs for derived classes. Not great because it's not generic, and only works if you have the concrete type anyway.
* Implement Read on a helper struct and allow conversion into that helper using `From`.
* Implement Read on a helper struct and add an explicit method to convert to it.

I went with the last option because converting with `From` and `Into` always seemed to unsightly type annotations, involving the concrete helper type. With a custom method, the helper type can be more anonymous.

Interestingly, the compiler required the `InputStreamRead` struct to be public when using the function from a different crate, *but it didn't require the calling code to be able to actually see it*. So now client code can call `into_read` and use the return value, but it can't name its concrete type. Which feels a little weird.


src/input_stream.rs
```diff
@@ -12,9 +12,11 @@ use glib::translate::*;
 use glib::Priority;
 use glib_ffi;
 use gobject_ffi;
+use std::io;
 use std::mem;
 use std::ptr;
 use InputStream;
+use super::error::to_std_io_result;
```
EPashkin: without `super::` it not works?

EPashkin:
Thanks, this part looks good for me

src/error.rs
```diff
@@ -0,0 +1,73 @@
+// Copyright 2018, The Gtk-rs Project Developers.
+// See the COPYRIGHT file at the top-level directory of this distribution.
+// Licensed under the MIT license, see the LICENSE file or <http://opensource.org/licenses/MIT>
+
```
GuillaumeGomez: Please remove extra line.

src/error.rs
```diff
@@ -0,0 +1,73 @@
...
+
+use Error;
+use IOErrorEnum;
+use std::io;
+
```
GuillaumeGomez: Please remove extra line.

src/error.rs
```diff
@@ -0,0 +1,73 @@
...
+
+impl From<IOErrorEnum> for io::ErrorKind {
+    fn from(kind: IOErrorEnum) -> Self {
+        match kind {
+            IOErrorEnum::Failed => io::ErrorKind::Other,
```
GuillaumeGomez: Since you return `io::ErrorKind::Other` a lot of times, why not grouping all error returning it?

src/input_stream.rs
```diff
@@ -204,9 +210,19 @@ impl<O: IsA<InputStream> + IsA<glib::Object> + Clone + 'static> InputStreamExtMa
     }
 }
 
+pub struct InputStreamRead<T: InputStreamExtManual>(T);
```
sdroege: Might be useful to also have a function to get back to the original type, and/or implement `Deref` for it

fkrull:
Ok, I added the std::io::Write part as well. If there's anything else holding up this PR, let me know.

src/error.rs
```diff
@@ -0,0 +1,36 @@
+// Copyright 2018, The Gtk-rs Project Developers.
+// See the COPYRIGHT file at the top-level directory of this distribution.
+// Licensed under the MIT license, see the LICENSE file or <http://opensource.org/licenses/MIT>
+
+use Error;
+use IOErrorEnum;
+use std::io;
+
+impl From<IOErrorEnum> for io::ErrorKind {
```
sdroege: This is generally very useful, thanks :)

sdroege:
Looks all good to me apart from the things already mentioned in the first comments above, which are mostly cosmetic :) Update that and we can get this in

fkrull:
> 
> 
> Looks all good to me apart from the things already mentioned in the first comments above, which are mostly cosmetic :) Update that and we can get this in

Oh, I fixed those earlier, I didn't add a comment for it though. Or is there anything else I missed?

sdroege:
> Oh, I fixed those earlier, I didn't add a comment for it though. Or is there anything else I missed?

Ah I just saw that the comments are still shown by github and it didn't detect them as "resolved". Will double check later :)

sdroege:
Indeed, looks all good. @GuillaumeGomez ?

src/input_stream.rs
sdroege: Should implement `Eq` and `Debug` here at least, based on the impls of the contained type. Can probably be derived

src/output_stream.rs
```diff
@@ -179,9 +186,30 @@ impl<O: IsA<OutputStream> + IsA<glib::Object> + Clone + 'static> OutputStreamExt
     }
 }
 
+pub struct OutputStreamWrite<T: OutputStreamExt>(T);
```
sdroege: Should implement `Eq` and `Debug` here at least, based on the impls of the contained type. Can probably be derived

GuillaumeGomez:
Looks good to me as well. Thanks!

sdroege:
@fkrull can you send a new PR that implements `Debug` and `PartialEq`and `Eq` for the new structs :)

sdroege:
And potentially something to come back from the new struct to the contained type again

PR #172, opened 2018-11-20
Add impls for new std::io::{Read,Write} structs

fkrull:
I added impls for Debug, PartialEq, Eq, Hash, Clone to the new helper structs.

@sdroege There already are `into_input_stream` and `into_output_stream` methods on the structs to turn them back into the gio types. Or did you mean something else?

### DIFF
--- a/src/input_stream.rs
+++ b/src/input_stream.rs
@@ -210,6 +210,7 @@ impl<O: IsA<InputStream> + IsA<glib::Object> + Clone + 'static> InputStreamExtMa
     }
 }
 
+#[derive(Debug, PartialEq, Eq, Hash, Clone)]
 pub struct InputStreamRead<T: InputStreamExtManual>(T);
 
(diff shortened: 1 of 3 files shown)


sdroege:
> @sdroege There already are `into_input_stream` and `into_output_stream` methods on the structs to turn them back into the gio types. Or did you mean something else?

@fkrull I was thinking of `Deref`, or a function to get the stream reference from it (without consuming the adapter). Do you think that would be useful?

Apart from that the PR here looks good. Sorry for not replying earlier!

fkrull:
I added some methods to get references to the wrapped streams. I decided against using `Deref` though since it seems semantically incorrect.

sdroege:
Looks good to me. @GuillaumeGomez ?

GuillaumeGomez:
Looks good to me as well. Thanks!

PR #175, opened 2018-12-01
Mark functions to create a socket from raw fd/socket as unsafe

sdroege:
It's unsafe as one could pass any possible number as fd/handle here from
safe code otherwise.

### DIFF
--- a/src/socket.rs
+++ b/src/socket.rs
@@ -14,6 +14,8 @@ use std::cell::RefCell;
 use std::mem::transmute;
 use Socket;
 use SocketAddress;
+#[cfg(all(not(windows), feature = "dox"))]
+use std::os::raw::c_void;
 #[cfg(all(not(unix), feature = "dox"))]
 use std::os::raw::c_int;
 use fragile::Fragile;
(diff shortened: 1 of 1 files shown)


EPashkin:
Good catch, thanks,
There still needed some work for `impl AsXX for Socket` generate doc for different system.

sdroege:
What work do you mean?

EPashkin:
Both CI failed on cargo doc --features "dox"

sdroege:
Ah didn't see that, I'm on the phone right now and the github mobile ui is quite useless for showing such things :)

Will fix later

sdroege:
Finally! All green @GuillaumeGomez :)

philn:
Was it green? `cargo build --features=futures,dox` fails here:

```
                                                                                                                                                                                                                  
error[E0605]: non-primitive cast: `i32` as `socket::RawSocket`                                                                                                                                                     
  --> src/socket.rs:64:13                                                                                                                                                                                          
   |                                                                                                                                                                                                               
64 |             ffi::g_socket_get_fd(self.to_glib_none().0) as _                                                                                                                                                  
   |             ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^                                                                                                                                                  
   |                                                                                                                                                                                                               
   = note: an `as` expression can only be used to convert between primitive types. Consider using the `From` trait       
```

sdroege:
Apparently it was, yes. But I can reproduce that.

PR #178, opened 2018-12-17
Fix socket.rs build with --features dox

sdroege:
error[E0605]: non-primitive cast: `i32` as `socket::RawSocket`
  --> src/socket.rs:64:13
   |
64 |             ffi::g_socket_get_fd(self.to_glib_none().0) as _
   |             ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
   |
   = note: an `as` expression can only be used to convert between primitive types. Consider using the `From` trait

----

See https://github.com/gtk-rs/gio/pull/175#issuecomment-447677702

### DIFF
--- a/src/socket.rs
+++ b/src/socket.rs
@@ -302,7 +302,7 @@ pub trait AsRawFd {
 }
 
 #[cfg(all(not(unix), feature = "dox"))]
-pub struct RawFd(c_int);
+pub type RawFd = c_int;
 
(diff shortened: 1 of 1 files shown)


GuillaumeGomez:
Thanks!

EPashkin:
@sdroege  Thanks, 👍

PR #179, opened 2019-01-15
Regen

sdroege:


EPashkin:
There still warning in manual code
```
warning: unused import: `glib::translate::*`
 --> D:/eap/rust/0/gio\src\subprocess_launcher.rs:6:5
  |
6 | use glib::translate::*;
  |     ^^^^^^^^^^^^^^^^^^
  |
  = note: #[warn(unused_imports)] on by default
```

sdroege:
> There still warning in manual code

I can't reproduce this here. How are you compiling?

EPashkin:
I build examples\pending with uncommented defaults.
Or just use features="v2_40"

sdroege:
>  Or just use features="v2_40"

Doesn't produce a warning for me either :( Can you fix it?

sdroege:
> Doesn't produce a warning for me either :( Can you fix it?

Also removing that line will cause compile errors for me when using v2_40

GuillaumeGomez:
Well then, I merge. :)

EPashkin:
Seems in only appeared for windows, so it need be something like `#[cfg(any(all(feature = "v2_40", unix), feature = "dox"))]`,
we can fix this after all merged

src/auto/file.rs
```diff
@@ -1429,44 +1393,42 @@ impl<O: IsA<File>> FileExt for O {
...
+        let callback = read_async_trampoline::<R>;
         unsafe {
-            ffi::g_file_read_async(self.to_glib_none().0, io_priority.to_glib(), cancellable.0, Some(callback), Box::into_raw(user_data) as *mut _);
+            ffi::g_file_read_async(self.as_ref().to_glib_none().0, io_priority.to_glib(), cancellable.map(|p| p.as_ref()).to_glib_none().0, Some(callback), Box::into_raw(user_data) as *mut _);
         }
     }
 
     #[cfg(feature = "futures")]
-    fn read_async_future(&self, io_priority: glib::Priority) -> Box_<futures_core::Future<Item = (Self, FileInputStream), Error = (Self, Error)>> where Self: Sized + Clone {
+    fn read_async_future<P: IsA<Cancellable> + Clone + 'static>(&self, io_priority: glib::Priority) -> Box_<futures_core::Future<Item = (Self, FileInputStream), Error = (Self, Error)>> where Self: Sized + Clone {
```
EPashkin: @sdroege Has trait constraint without usage :(
so examples still failing.

EPashkin: Added issue to rust: https://github.com/rust-lang/rust/issues/57660

sdroege: Yeah but that's not a problem by itself, only here we have the constraint but shouldn't have it :)

PR #192, opened 2019-02-08
Bring Travis and AppVeyor configurations in sync with the GLib one

sdroege:


.travis.yml
```diff
@@ -50,14 +50,12 @@ before_install:
     fi
 script:
   - rustc --version
-  - if [ "$TRAVIS_RUST_VERSION" == "nightly" ] && [ "$GTK" == "3.4" ]; then
+  - if [ "$TRAVIS_RUST_VERSION" == "nightly" ] && [ "$GTK" == "3.14" ]; then
     make regen_check;
     fi
-  - cargo test --features "embed-lgpl-docs v2_40" --jobs 1
   - cargo doc --features "dox"
-  # catch any sneaked in lgpl docs
```
sdroege: Note that the lgpl-docs stuff is not done here anymore as it's also not done inside GLib. I can add it back but then I'd also would have to add it to GLib. I'd like to have all these to stay in sync.

Btw a nice feature of gitlab ci is that you can have template CI configurations, and then only override the things that you need in the specific repos/jobs. That would help here a lot.

.travis.yml
```diff
@@ -1,43 +1,43 @@
+dist: xenial
 language: rust
 matrix:
   include:
     - os: linux
       rust: nightly
-      env: GTK=3.4
+      env: GTK=3.14 FEATURES=
```
GuillaumeGomez: Isn't something missing after `FEATURES=`?

sdroege: it's empty because we build with the default features :)

GuillaumeGomez: Ok!

GuillaumeGomez:
👍

.travis.yml
EPashkin: `cargo test --features "$FEATURES""` looks strange but seems cargo allow it

EPashkin:
👍

PR #199, opened 2019-02-24
Regen

GuillaumeGomez:
cc @EPashkin @sdroege


src/auto/application_command_line.rs
```diff
@@ -145,7 +145,7 @@ impl<O: IsA<ApplicationCommandLine>> ApplicationCommandLineExt for O {
 
 unsafe extern "C" fn notify_is_remote_trampoline<P, F: Fn(&P) + 'static>(this: *mut ffi::GApplicationCommandLine, _param_spec: glib_ffi::gpointer, f: glib_ffi::gpointer)
 where P: IsA<ApplicationCommandLine> {
-    let f: &F = transmute(f);
+    let f: &F = &*(f as *const F);
```
EPashkin: @GuillaumeGomez this change also better applied to manual code, but I don't insist

EPashkin: with removing `#[cfg_attr(feature = "cargo-clippy", allow(transmute_ptr_to_ref))]` if there present

EPashkin:
👍

src/auto/application_command_line.rs
GuillaumeGomez: Updating.

GuillaumeGomez: This is the only actual case where the `transmute` can be easily changed without having a huge (and ugly) `as stuff`.

sdroege:
:+1:

PR #201, opened 2019-03-09
inet_address: add new_from_bytes constructor

abdulrehman-git:


GuillaumeGomez:
Looks good, thanks!

cc @sdroege @EPashkin

EPashkin:
@abdulrehman-git Thanks
 👍

src/lib.rs
```diff
@@ -49,6 +49,7 @@ mod subprocess;
 mod subprocess_launcher;
 #[cfg(any(unix, feature = "dox"))]
 mod unix_socket_address;
+mod inet_address;
```
sdroege: Don't we need a `pub use inet_address::InetAddressOctet` here then?

src/inet_address.rs
```diff
@@ -0,0 +1,40 @@
+use SocketFamily;
+use ffi;
+use glib::translate::*;
+use InetAddress;
+
+#[derive(Debug)]
+pub enum InetAddressOctet<'a> {
```
sdroege: The name does not seem ideal. It's not a single octet (== byte) but 4 or 16 bytes. Maybe instead of using the rare word octet, we could call this `InetAddressBytes` instead?

abdulrehman-git: Okay, let me rename :-), I used `Octet` to match `ip.octets()`.

src/lib.rs
abdulrehman-git: Ah, let me fix.

src/inet_address.rs
sdroege: @EPashkin @GuillaumeGomez do you have suggestions? I would call it bytes because that's how it's called in GIO :)

EPashkin: IMHO `Octets` or `Bytes` is almost same, `Octet` is really not.
https://en.wikipedia.org/wiki/IP_address uses "group of 8 bits (an octet)" but not `byte` maybe because byte can have different size.
So I for `InetAddressOctets`

GuillaumeGomez: In french, octet is the translation of byte so it seemed good to me. :3

For the name, `InetAddressData` might be enough?

sdroege: @EPashkin That was true decades ago, nowadays a byte is always 8 bit fortunately. And the GIO API talks about bytes here (`new_from_bytes()` for example), not octets.

EPashkin: Ok, then `Bytes`, just `Data` IMHO not clear

abdulrehman-git: I had force-pushed with InetAddressBytes :-).

PR #203, opened 2019-03-10
Install missing dependency

GuillaumeGomez:


EPashkin:
I already tried that, it already present in system,
see: https://github.com/gtk-rs/gtk/pull/795#issuecomment-470670457

.travis.yml
```diff
@@ -45,7 +45,7 @@ addons:
     - libmount-dev
 before_install:
   - if [[ "$TRAVIS_OS_NAME" == "osx" ]]; then brew update; fi
-  - if [[ "$TRAVIS_OS_NAME" == "osx" ]]; then brew install gtk+3 cairo atk; fi
+  - if [[ "$TRAVIS_OS_NAME" == "osx" ]]; then brew install libffi gtk+3 cairo atk; fi
```
sdroege: libffi is a dependency of glib, so this is only a workaround for brew being broken. Can you add a comment about that here?

GuillaumeGomez: Sure!

GuillaumeGomez:
@EPashkin Damn, I completely forgot about it... Do you know if there is an open issue somewhere about this?

EPashkin:
IMHO it apveyor image problem, it need be refreshed.
Fast search did not show same problem 2 days ago or any way disable brew's autocleanup

PR #206, opened 2019-04-26
Port to futures 0.3

sdroege:
While this requires nightly, the API is now stabilized and will become part of Rust 1.36. IMHO we should merge this now so that it can all get wider testing.


sdroege:
This is currently marked as WIP because it uses https://github.com/gtk-rs/glib/pull/476 and https://github.com/gtk-rs/gir/pull/753 via Cargo patching

sdroege:
I'll also change this to actually run the futures tests on nightly. And send another PR for that for glib.

Otherwise please review this one already but not merge yet.

sdroege:
> I'll also change this to actually run the futures tests on nightly. And send another PR for that for glib.

Added and done

EPashkin:
@sdroege Thanks. 👍.
Sourceview's `auto/search_context.rs` also affected by gir changes

sdroege:
So let's get this in and then I do the examples first (they don't need the other crates updated), and then I do gdk-pixbuf, gtk and sourceview.

PR #208, opened 2019-05-06
 Add Gir config for Unix{Input,Output}Stream

moggiesir:
I put auto/* files into a separate commit as the docs on gtk-rs/gtk said.

Not sure if there's something to watch for in the generated output, but this worked for connecting a WebKit2.URISchemeRequest.finish to a either a socketpair or unix pipe.


GuillaumeGomez:
:+1:

src/auto/unix_input_stream.rs
```diff
@@ -0,0 +1,84 @@
...
+glib_wrapper! {
+    pub struct UnixInputStream(Object<gio_sys::GUnixInputStream, gio_sys::GUnixInputStreamClass, UnixInputStreamClass>) @extends InputStream, @implements PollableInputStream;
+
+    match fn {
+        get_type => || gio_sys::g_unix_input_stream_get_type(),
+    }
+}
+
+impl UnixInputStream {
+    pub fn new(fd: i32, close_fd: bool) -> UnixInputStream {
```
sdroege: This has to be via `IntoRawFd` and `unsafe` instead of just taking a random integer as argument for the file descriptor. See `gio::Socket` bindings.

src/auto/unix_input_stream.rs
```diff
@@ -0,0 +1,84 @@
...
+        }
+    }
+}
+
+pub const NONE_UNIX_INPUT_STREAM: Option<&UnixInputStream> = None;
+
+pub trait UnixInputStreamExt: 'static {
+    fn get_close_fd(&self) -> bool;
+
+    fn get_fd(&self) -> i32;
```
sdroege: Via `FromRawFd` (and does not need to be unsafe). You probably also want to implement `AsRawFd` on both streams.

src/auto/unix_output_stream.rs
```diff
@@ -0,0 +1,84 @@
...
+glib_wrapper! {
+    pub struct UnixOutputStream(Object<gio_sys::GUnixOutputStream, gio_sys::GUnixOutputStreamClass, UnixOutputStreamClass>) @extends OutputStream, @implements PollableOutputStream;
+
+    match fn {
+        get_type => || gio_sys::g_unix_output_stream_get_type(),
+    }
+}
+
+impl UnixOutputStream {
+    pub fn new(fd: i32, close_fd: bool) -> UnixOutputStream {
```
sdroege: See above, `IntoRawFd` and `unsafe`

src/auto/unix_output_stream.rs
```diff
@@ -0,0 +1,84 @@
...
+        }
+    }
+}
+
+pub const NONE_UNIX_OUTPUT_STREAM: Option<&UnixOutputStream> = None;
+
+pub trait UnixOutputStreamExt: 'static {
+    fn get_close_fd(&self) -> bool;
+
+    fn get_fd(&self) -> i32;
```
sdroege: See above, `FromRawFd` and possibly `AsRawFd` impl on the stream

src/auto/unix_input_stream.rs
moggiesir: Could you explain why this needs to be `unsafe` if it takes a `IntoRawFd`?

AFAICT, `IntoRawFd` is defined as taking `self` (without &), so the call to `into_raw_fd` should move the src, meaning nothing else can use the fd at that point, right? Is there something else about the call that is unsafe?

I feel like to satisfy Rust's safety constraints for RawFd using classes, the close_fd parameter doesn't really make sense to expose this way. Maybe via a different function that takes an AsRawFd.

src/auto/unix_input_stream.rs
moggiesir: To satisfy the FromRawFd docs, get_fd should probably be defined as moving self so that the UnixInputStream is no longer usable, right?

src/auto/unix_input_stream.rs
sdroege: You can do the following with completely safe code to generate invalid fds from random numbers:

```rust
struct Meh(i32);

impl IntoRawFd for Meh {
  fn into_raw_fd(self) -> RawFd { self.0 }
}

let fd: RawFd = Meh(123).into_raw_fd();
```

So any function taking a `RawFd` must be marked as unsafe as the caller need to ensure that the passed value is valid and ok to use in this context.


For `close_fd`, I agree.

src/auto/unix_input_stream.rs
sdroege: The problem here is that all `glib::Object`s are reference counted. So even if you take self by value, another reference could still be around elsewhere.

For `FromRawFd`, I meant an implementation like the one for the socket: https://github.com/gtk-rs/gio/blob/047ea1cf0735f04c5a404331c61d192ac4cb43a9/src/socket.rs#L197-L202

sdroege: And note that this is (probably?) fine as every API doing something with raw fds must be unsafe (see above), so whenever dealing with raw fds it's the programmer's job to ensure that things are done correctly.

moggiesir: Done

sdroege:
lgtm, thanks!

@GuillaumeGomez ?

GuillaumeGomez:
Looks good to me as well, thanks! (and big thanks for the comments in the gir file!)

PR #209, opened 2019-05-11
Add support for subclassing gio::Application

sdroege:
Based on code by Dirk Van Haerenborgh @vhdirk from https://github.com/gtk-rs/gobject-subclass

### DIFF
--- a/src/subclass/application.rs
+++ b/src/subclass/application.rs
@@ -1,0 +1,529 @@
+// Copyright 2019, The Gtk-rs Project Developers.
+// See the COPYRIGHT file at the top-level directory of this distribution.
+// Licensed under the MIT license, see the LICENSE file or <http://opensource.org/licenses/MIT>
+
+use gio_sys;
+use glib_sys;
+
+use glib::translate::*;
+
+use glib::subclass::prelude::*;
+
+use Application;
+use ApplicationClass;
+
+use libc::{c_char, c_void};
+use std::convert;
+use std::ffi::OsString;
+use std::fmt;
+use std::ops::Deref;
+use std::ptr;
+
+pub struct ArgumentList {
+    pub(crate) ptr: *mut *mut *mut c_char,
+    items: Vec<OsString>,
+}
+
+impl ArgumentList {
+    pub(crate) fn new(arguments: *mut *mut *mut c_char) -> Self {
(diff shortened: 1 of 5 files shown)


src/subclass/application.rs
```diff
@@ -0,0 +1,529 @@
+// Copyright 2019, The Gtk-rs Project Developers.
+// See the COPYRIGHT file at the top-level directory of this distribution.
+// Licensed under the MIT license, see the LICENSE file or <http://opensource.org/licenses/MIT>
+use gio_sys;
```
GuillaumeGomez: You have the right to add an empty line before this one. :)

sdroege: done

src/subclass/application.rs
```diff
@@ -0,0 +1,530 @@
...
+use ApplicationClass;
+
+use libc::{c_char, c_void};
+use std::convert;
+use std::ffi::OsString;
+use std::fmt;
+use std::ops::Deref;
+use std::ptr;
+
+pub struct ArgumentList {
```
EPashkin: Why not move `ArgumentList` to separate file?

src/subclass/application.rs
```diff
@@ -0,0 +1,530 @@
...
+
+impl fmt::Debug for ArgumentList {
+    fn fmt(&self, formatter: &mut fmt::Formatter) -> Result<(), fmt::Error> {
+        self.items.fmt(formatter)
+    }
+}
+
+impl convert::Into<Vec<OsString>> for ArgumentList {
+    fn into(self) -> Vec<OsString> {
+        self.items.clone()
```
EPashkin: Clone really needed? Why not replace it with empty vector?

EPashkin:
@sdroege Thanks

src/subclass/application.rs
sdroege: Because it will only ever be used inside this file

src/subclass/application.rs
sdroege: Indeed it should be possible to just move it out

sdroege: Fixed. Thanks :)

EPashkin:
@sdroege 👍

PR #213, opened 2019-05-18
Generate new types

GuillaumeGomez:
Needs to be regenerated after https://github.com/gtk-rs/gir/pull/774 and https://github.com/gtk-rs/glib/pull/486

cc @sdroege @EPashkin


GuillaumeGomez:
Now just remains gtk-rs/glib#486 and we're good to go.

src/auto/file.rs
```diff
@@ -205,12 +205,12 @@ pub trait FileExt: 'static {
 
     fn make_symbolic_link<P: AsRef<std::path::Path>, Q: IsA<Cancellable>>(&self, symlink_value: P, cancellable: Option<&Q>) -> Result<(), Error>;
 
-    //fn measure_disk_usage<P: IsA<Cancellable>>(&self, flags: /*Ignored*/FileMeasureFlags, cancellable: Option<&P>, progress_callback: Option<Box<dyn Fn(bool, u64, u64, u64) + 'static>>) -> Result<(u64, u64, u64), Error>;
+    fn measure_disk_usage<P: IsA<Cancellable>>(&self, flags: FileMeasureFlags, cancellable: Option<&P>, progress_callback: Option<Box<dyn Fn(bool, u64, u64, u64) + 'static>>) -> Result<(u64, u64, u64), Error>;
 
-    //fn measure_disk_usage_async<P: IsA<Cancellable>, Q: FnOnce(Result<(u64, u64, u64), Error>) + Send + 'static>(&self, flags: /*Ignored*/FileMeasureFlags, io_priority: glib::Priority, cancellable: Option<&P>, progress_callback: Option<Box<dyn Fn(bool, u64, u64, u64) + 'static>>, callback: Q);
+    //fn measure_disk_usage_async<P: IsA<Cancellable>, Q: FnOnce(Result<(u64, u64, u64), Error>) + Send + 'static>(&self, flags: FileMeasureFlags, io_priority: glib::Priority, cancellable: Option<&P>, progress_callback: Option<Box<dyn Fn(bool, u64, u64, u64) + 'static>>, callback: Q);
```
sdroege: Why is this not generated still? Seems like nothing is actually missing now

src/auto/file_attribute_matcher.rs
```diff
@@ -0,0 +1,70 @@
...
+        }
+    }
+
+    pub fn enumerate_namespace(&self, ns: &str) -> bool {
+        unsafe {
+            from_glib(gio_sys::g_file_attribute_matcher_enumerate_namespace(self.to_glib_none().0, ns.to_glib_none().0))
+        }
+    }
+
+    pub fn enumerate_next(&self) -> Option<GString> {
```
sdroege: This looks like you'd want to implement an `Iterator` here :) Please add an `iter()` method that returns a custom `Iterator` around this function.

sdroege: Or rather `into_iter()` as the matcher is only really usable once.

src/auto/file_attribute_matcher.rs
```diff
@@ -0,0 +1,70 @@
...
+        }
+    }
+
+    pub fn matches_only(&self, attribute: &str) -> bool {
+        unsafe {
+            from_glib(gio_sys::g_file_attribute_matcher_matches_only(self.to_glib_none().0, attribute.to_glib_none().0))
+        }
+    }
+
+    pub fn subtract(&self, subtract: &FileAttributeMatcher) -> Option<FileAttributeMatcher> {
```
sdroege: I don't think this can return `None`

src/auto/file_info.rs
```diff
@@ -233,7 +246,7 @@ impl FileInfo {
         }
     }
 
-    //pub fn set_attribute(&self, attribute: &str, type_: /*Ignored*/FileAttributeType, value_p: /*Unimplemented*/Fundamental: Pointer) {
+    //pub fn set_attribute(&self, attribute: &str, type_: FileAttributeType, value_p: /*Unimplemented*/Fundamental: Pointer) {
```
sdroege: Seems like we want some manual implementation for all this file attribute API, that safely exposes this pointer via some trait with an associated type for the different file attribute types

src/auto/settings.rs
```diff
@@ -83,7 +83,7 @@ pub trait SettingsExt: 'static {
 
     fn bind<P: IsA<glib::Object>>(&self, key: &str, object: &P, property: &str, flags: SettingsBindFlags);
 
-    //fn bind_with_mapping<P: IsA<glib::Object>>(&self, key: &str, object: &P, property: &str, flags: SettingsBindFlags, get_mapping: /*Unimplemented*/Fn(/*Ignored*/glib::Value, &glib::Variant) -> bool, set_mapping: /*Unimplemented*/Fn(/*Ignored*/glib::Value, &glib::VariantType) -> glib::Variant, user_data: /*Unimplemented*/Option<Fundamental: Pointer>);
+    //fn bind_with_mapping<P: IsA<glib::Object>, Q: Fn(&glib::Value, &glib::Variant) -> bool + 'static, R: Fn(&glib::Value, &glib::VariantType) -> glib::Variant + 'static>(&self, key: &str, object: &P, property: &str, flags: SettingsBindFlags, get_mapping: Q, set_mapping: R);
```
sdroege: Why is this not generated? Seems complete now

src/auto/file.rs
GuillaumeGomez: I wonder...

src/auto/file_attribute_matcher.rs
GuillaumeGomez: Yes sir!

src/auto/file_attribute_matcher.rs
GuillaumeGomez: I'll check.

src/auto/settings.rs
GuillaumeGomez: I don't know either. I'll need to check gir output.

src/auto/file.rs
GuillaumeGomez: `measure_disk_usage_async: missing success parameters for async future`

src/auto/settings.rs
GuillaumeGomez: It doesn't detect the user data apparently: "function `g_settings_bind_with_mapping`'s callback `get_mapping` without associated user data"

src/auto/file_attribute_matcher.rs
GuillaumeGomez: It seems like it can: https://gitlab.gnome.org/GNOME/glib/blob/master/gio/gfileinfo.c#L2388

src/auto/file.rs
sdroege: What does that mean? :)

src/auto/settings.rs
sdroege: Can you create an issue for gir about this?

src/auto/file_attribute_matcher.rs
sdroege: Indeed, and it's valid to pass `NULL` too for both parameters

src/auto/file.rs
GuillaumeGomez: Aren't you the one who wrote this part of gir? :p Or maybe @antoyo?

src/auto/settings.rs
GuillaumeGomez: I need to check if it comes from the gir file first.

src/auto/file_attribute_matcher.rs
GuillaumeGomez: Ah! I need to make (at least) the second one nullable then.

src/auto/file.rs
sdroege: Not me :)

antoyo: @GuillaumeGomez: Looks like [you](https://github.com/gtk-rs/gir/blame/56f6de49e185fb3ff8502ed0d4ddf447d8d5fc05/src/analysis/functions.rs#L725) did it.

GuillaumeGomez: Ok, maybe it's on me.... Again... XD

src/auto/settings.rs
GuillaumeGomez: Seems like it comes from gir... Opening an issue.

GuillaumeGomez:
Updated.

src/auto/file.rs
sdroege: Let's create a gir issue for this then? :)

src/auto/file_info.rs
sdroege: For a later PR

sdroege: Just keep in mind that the API is kind of useless without

sdroege:
:+1:

src/auto/file.rs
GuillaumeGomez: https://github.com/gtk-rs/gir/issues/783

src/auto/file_info.rs
GuillaumeGomez: I'll try to add it as soon as possible.

sdroege: I think the nice way to implement would be similar to the GStreamer tags support. You'd have a ZST for each attribute that implements a trait, that trait has an associated type for the attribute type and provides ways for getting from such a value to the attribute

PR #218, opened 2019-06-08
Move trampolines

GuillaumeGomez:
Part of gtk-rs/gir#792


sdroege:
Will you also do GLib? In object.rs there should be at least one, for the notify signal.

GuillaumeGomez:
When I generated, nothing changed?

sdroege:
There's manual code :)

GuillaumeGomez:
Oh you meant updating the manual code! In case a function is used in different places, what should we do? Copy it or keep it "outside"?

sdroege:
Not copy :) Keep it then

PR #219, opened 2019-06-12
Format code with rustfmt

sdroege:


Makefile
```diff
@@ -14,6 +15,7 @@ not_bound: $(GIR) $(GIR_FILES)
 regen_check: $(GIR) $(GIR_FILES)
 	rm src/auto/*
 	$(GIR) -c Gir.toml
+	cargo fmt
```
EPashkin: Maybe simpler add this to gir?

sdroege: And you're saying that after that same change was merged into glib :)

IMHO it's nicer to keep this separate from gir: that way it's a decision of the caller what to do and when. And putting it into gir would make gir fail if there's no rustfmt available (see the other change here to add it in .travis.yml).

GuillaumeGomez: I also prefer to keep it separated from gir.

EPashkin: Ok

Issue #161, opened 2018-10-18
More functions that probably shouldn't return an Option


sophie-h:
- `gio::File::get_uri()` https://gitlab.gnome.org/GNOME/glib/blob/master/gio/gfile.c#L639
- `gio::File::get_uri_scheme()` https://gitlab.gnome.org/GNOME/glib/blob/master/gio/gfile.c#L449

and

```c
#define G_IS_FILE(obj)	       (G_TYPE_CHECK_INSTANCE_TYPE ((obj), G_TYPE_FILE))
```

If I am correct, there are probably many more functions, that shouldn't return an `Option`.

EPashkin:
By https://gitlab.gnome.org/GNOME/glib/blob/master/gio/gfile.c#L6835 `get_uri_scheme` can return  NULL in some cases.
At minimum `get_uri` implementation return NULL for nonabsolute paths
https://gitlab.gnome.org/GNOME/glib/blob/master/gio/glocalfile.c#L277
https://gitlab.gnome.org/GNOME/glib/blob/master/glib/gconvert.c#L1774

sophie-h:
I think `GFile` has no concept/representation of relative paths. Relative arguments to `g_file_new_for_path()` lead to an absolute representation in `g_file_get_uri()`. Also, NULL is not documented as return value.

EPashkin:
You right, for relative paths GLocalFile doing canonicalize.
I found only one call in gio, that checks it result for NULL: https://gitlab.gnome.org/GNOME/glib/blob/master/gio/win32/gwinhttpfile.c#L243,
so maybe you right about `get_uri`

sdroege:
So anything that has to be done here?

sophie-h:
@sdroege it's probably more the question where to start? There is a lot of functions that don't need to return an Option. I don't know if it is really efficient if I am opening an issue for every singe function (#158, gtk-rs/gtk#709).

sdroege:
Let's close this then. Please open new issues for everything you notice that shouldn't return an `Option` :)

sophie-h:
> The functions mentioned here seem to be correct though? Unlikely that they return `None` but unfortunately that can

I don't think so.

sdroege:
Checked again and you're right that these functions shouldn't return an `Option`. The discussion above confused me! Will fix :)

sdroege:
See https://github.com/gtk-rs/gio/pull/221

Issue #99, opened 2018-04-02
SocketListener::accept_async panics


antoyo:
Hi.
The current `SocketListener::accept_async` method does not work because the `Object` parameter of the callback should be optional (and the `.gir` file says so) but it is not.
It might be because `gir` does not look at the `optional` attribute when generating the callback.

EPashkin:
Gir don't look at this attribute at all, only at `allow-none` but don't trusts him too, as many cases when it wrong.
In current realization of async trampoline IMHO will be touch adding standard config checking as trampoline not analyzed.

sdroege:
For the async trampoline we should maybe reverse the rules and assume by default that everything can be NULL unless overridden.

The problem basically is that GIR annotations exist only for "this can be NULL", but it can be easily forgotten in C libraries so the absence of that attribute does not necessarily mean that it must not be NULL / can't be NULL. So for return values we always assume that they can be NULL unless overridden in the toml file, for arguments we assume they must not be NULL unless the attribute is set (or overridden in the toml file).
What would help is a "this can not be NULL" annotation for return values.

sdroege:
For now maybe just implement this specific function manually?

sdroege:
@antoyo What's the status here? Is there still a problem, and if so can you provide a testcase to reproduce it?

sdroege:
It is. Fixing

sdroege:
See https://github.com/gtk-rs/gio/pull/221

PR #221, opened 2019-06-19
Two Option-related fixes

sdroege:
See https://github.com/gtk-rs/gio/issues/99 and https://github.com/gtk-rs/gio/issues/161


GuillaumeGomez:
:+1:

PR #226, opened 2019-07-05
settings: add SettingsExtManual Trait for get::<T>/set::<T> fn

Cogitri:


Cogitri:
This current throws
```
error[E0308]: mismatched types
  --> src/settings.rs:15:31
   |
15 |         self.set_boolean(key, value)
   |                               ^^^^^ expected bool, found type parameter
   |
   = note: expected type `bool` (bool)
              found type `bool` (type parameter)
```

Not quite positive what to make of that.

I'll add more types once this works.

EPashkin:
This because you added `bool` as implementation constraint:
```
impl<bool, O: SettingsExt> SettingsExtManual<bool> for O {
```
try just
```
impl<O: SettingsExt> SettingsExtManual<bool> for O {
```

src/prelude.rs
```diff
@@ -15,11 +15,12 @@ pub use list_store::ListStoreExtManual;
...
+pub use settings::SettingsExtManual;
 pub use socket::*;
 pub use socket_listener::SocketListenerExtManual;
 #[cfg(any(unix, feature = "dox"))]
 pub use unix_input_stream::UnixInputStreamExtManual;
 #[cfg(any(unix, feature = "dox"))]
 pub use unix_output_stream::UnixOutputStreamExtManual;
 #[cfg(any(unix, feature = "dox"))]
-pub use unix_socket_address::{UnixSocketAddressExtManual, UnixSocketAddressPath};
+pub use unix_socket_address::{UnixSocketAddressExtManual, UnixSocketAddressPath};
```
EPashkin: This change cause CI fail on regen_check step

sdroege: Yes, it needs a `cargo fmt` run

src/settings.rs
```diff
@@ -0,0 +1,17 @@
+use SettingsExt;
+
+pub trait SettingsExtManual<T> {
+    fn get(&self, key: &str) -> T;
+
+    fn set(&self, key: &str, value: T) -> bool;
+}
+
+impl<bool, O: SettingsExt> SettingsExtManual<bool> for O {
```
sdroege: You need a custom trait here for the type, something like

```rust
pub trait SettingsValueType {
    fn get<T: IsA<Settings>>(&T) -> Self;
    fn set<T: IsA<Settings>>(&T, &self);
}

impl SettingsValueType for bool {
    [...]
}

impl<O: SettingsExt> SettingsExtManual for O {
    fn get<T: SettingsValueType>(&self, key: &str) -> T {
        <T as SettingsValueType>::get(self)
    }

    fn set<T: SettingsValueType>(&self, key: &str, value: &T) {
        <T as SettingsValueType>::set(self, value)
    }
}
```

Cogitri: Oh, alright, thanks

src/prelude.rs
Cogitri: Ah, normally CLion runs that for me, but it's a per-project setting I forgot to check here

src/settings.rs
Cogitri: Done

src/settings.rs
```diff
@@ -0,0 +1,138 @@
+use glib::{GString, Variant};
+use SettingsExt;
+
+pub trait SettingsGetValueType {
+    fn get<T: SettingsExt>(&T, key: &str) -> Self;
+}
+
+pub trait SettingsSetValueType {
+    fn set<T: SettingsExt>(&T, key: &str, val: Self) -> bool;
+}
```
Cogitri: I've split this into two Traits, since sometimes we return a different value in `set_` than we take in `get_` (e.g. with `get/set_string`, we take `&str` in `set_string` but return `Option<GString>` in `get_string`

sdroege:
Instead of introducing a new trait here, I'd suggest that we fix up the `GVariant` API so it works closer to how `GValue` works. And then you can use exactly the same traits (`ToVariant` / `FromVariant`) from `GVariant` here together with `g_settings_get_value()` / `g_settings_set_value()`.

Makes no sense to replicate the `GVariant` machinery here another time.

Cogitri:
Ah, so you want to just always use `get_value` (and converting to the requested type)/`set_value` (and converting the type to a Variant)? What exactly would need to be fixed in `Variant`?

sdroege:
Maybe nothing, is anything missing in `GVariant` for implementing this here?

In any case, that would be the easiest and in the end equivalent to calling `get_boolean()`, etc. anyway.

Cogitri:
> Maybe nothing, is anything missing in GVariant for implementing this here?

Oh, alright, ` I'd suggest that we fix up the GVariant API ` sounded to me as if something needed to be changed :)

Will work on it tomorrow, thanks for the pointer.

Cogitri:
Alright, this now uses the ToVariant/FromVariant API

src/settings.rs
```diff
@@ -0,0 +1,20 @@
...
+
+impl<O: IsA<Settings>> SettingsExtManual for O {
+    fn get<U: FromVariant>(&self, key: &str) -> Option<U> {
+        self.get_value(key)
+            .and_then(|val| FromVariant::from_variant(&val))
+    }
+
+    fn set<U: ToVariant>(&self, key: &str, value: &U) -> bool {
+        self.set_value(key, &value.to_variant())
```
sdroege: This `bool` should probably be a `Result<(), glib::BoolError>` :) Probably also for any other `set()` functions on `Settings`

src/settings.rs
```diff
@@ -0,0 +1,20 @@
...
+pub trait SettingsExtManual {
+    fn get<U: FromVariant>(&self, key: &str) -> Option<U>;
+
+    fn set<U: ToVariant>(&self, key: &str, value: &U) -> bool;
+}
+
+impl<O: IsA<Settings>> SettingsExtManual for O {
+    fn get<U: FromVariant>(&self, key: &str) -> Option<U> {
+        self.get_value(key)
+            .and_then(|val| FromVariant::from_variant(&val))
```
sdroege: Should probably return some kind of `Result` with a meaningful error type to distinguish "key not found" and "value of wrong type"

sdroege:
> Alright, this now uses the ToVariant/FromVariant API

Thanks, looks much nicer now :)

src/settings.rs
Cogitri: Hm, if you think that's in the scope of this PR, sure. Should I make wrapper functions for set_* or move the entirety of auto/settings.rs's set_* into this?

src/settings.rs
Cogitri: Alright

src/settings.rs
sdroege: You can tell the code generator, see https://gitlab.freedesktop.org/gstreamer/gstreamer-rs/blob/master/Gir_Gst.toml#L1291

It's an easy change so I thought it would be good to have as part of this PR

src/settings.rs
sdroege: See the (new since yesterday) GValue API for some inspiration (or copy&paste)

src/settings.rs
Cogitri: Oh, alright, hadn't known about that. Thanks for the pointer :)

Cogitri:
Alright, added `Result`s, will write some tests now to see it in action.

Gir.toml
```diff
@@ -605,6 +605,10 @@ status = "generate"
     [[object.signal]]
     name = "writable-change-event"
     inhibit = true
+    [[object.function]]
+    name="set_*"
+        [object.function.return]
+        bool_return_is_error = "Can't set readonly key"
```
sdroege: This doesn't return `false` if the key was of the wrong type? What happens if you try to set a boolean on an integer key?

Cogitri: I'm pretty sure GLib just SIGTRAPs then

Cogitri: And the docs say FALSE == not writeable, so I interpreted that as readonly: https://developer.gnome.org/gio/stable/GSettings.html#g-settings-set-int

Cogitri: Yup,
```
(process:13224): GLib-GIO-ERROR **: 23:42:07.005: Settings schema 'com.github.gtk-rs.test' does not contain a key named 'no-such-key'
test settings::test::bool_set_get ... ok
error: process didn't exit successfully: `/home/rasmus/Projects/gio/target/debug/deps/gio-8e55b7fc5fe6fb99` (signal: 5, SIGTRAP: trace/breakpoint trap)
```

Cogitri: Alright, at this point I'm a bit unsure why we even return Option<Variant> from SettingsExt::get_value()

Cogitri:
Alright, I guess the new version is more in line what GLib expects us to do, since it expects getting an unknown key to be a programmer error. I've made fixup commits so you can see the difference.

Gir.toml
sdroege: `get_value()` simply explodes too if the key does not exist? I.e. it can't possibly return `None`? In that case it should indeed not return an `Option` and this has to be fixed here too, leaving only the type mismatch as a valid error.

If you call `get_string()` on a bool setting, it also explodes? In that case all these functions should simply return their value and all the error handling added is not actually useful because of how the GLib API works. You could then simply panic in the getter if the type is wrong as that's consistent with how the GLib API works.

Sorry for the confusion but I wasn't aware of this behaviour of the GLib API, that it simply kills the process if the type is wrong or the key does not exist. Not very user-friendly API.

Cogitri: > If you call get_string() on a bool setting, it also explodes? In that case all these functions should simply return their value and all the error handling added is not actually useful because of how the GLib API works. You could then simply panic in the getter if the type is wrong as that's consistent with how the GLib API works.

Hm, I don't see a problem with keeping the type mismatch error since we can still handle that (since we just get the Value and convert it ourselves). The default GLib expectation _is_ that it's a programmer error if you don't know what kind of key you're getting though, so we can go with that too if you want.
 
> Sorry for the confusion but I wasn't aware of this behaviour of the GLib API, that it simply kills the process if the type is wrong or the key does not exist.

No worries :)

> Not very user-friendly API.

Well, usually it should be easy enough to guarentee that your GSchema is alright, the problem is when one wants to fetch some settings from foreign schemas (e.g. from the schema contained in gsettings-desktop-schemas, which are _always_ a pain to upgrade for distros because you have to bump the whole desktop at once or everything just SIGTRAPs because some key has been changed :/). As such I feel like we should keep at least the type error, the user can always just `unwrap()` if he wants to. I suppose we could even implement the `No such key` error by checking the key against `SettingsExt::list_keys()`, but that is potentially too expensive and I don't think list_keys is really meant for that purpose.

sdroege: > Hm, I don't see a problem with keeping the type mismatch error since we can still handle that (since we just get the Value and convert it ourselves). The default GLib expectation is that it's a programmer error if you don't know what kind of key you're getting though, so we can go with that too if you want.

In all other functions a missing key or type mismatch causes the process to be killed be GLib. So we should try to be consistent with that and at least panic :)

> I suppose we could even implement the No such key error by checking the key against SettingsExt::list_keys(), but that is potentially too expensive and I don't think list_keys is really meant for that purpose.

Yes, that's something an application could do. We shouldn't diverge that much from the C API.

Cogitri: Alright, sure 👍

Cogitri: Panicking here now

build.rs
```diff
@@ -15,3 +21,29 @@ fn manage_docs() {
 
 #[cfg(not(any(feature = "embed-lgpl-docs", feature = "purge-lgpl-docs")))]
 fn manage_docs() {}
+
+#[cfg(feature = "test")]
+fn generate_tests() {
```
Cogitri: I don't think cfg(test) nor this actually works in build.rs ...

Cogitri:
Other than that I think this is good to go now (?)

build.rs
sdroege: It doesn't, known cargo problem

sdroege:
> Other than that I think this is good to go now (?)

Indeed, thanks for working on this :)

build.rs
Cogitri: So, what can we do about this? Do we have to generate the gschema even for normal builds, or should we generate it in Settings' test module?

Cogitri:
> Indeed, thanks for working on this :)

Well, thanks for all the pointers! :)

build.rs
sdroege: Part of the test I guess

build.rs
```diff
@@ -15,3 +21,29 @@ fn manage_docs() {
...
+    let output = Command::new("glib-compile-schemas")
```
sdroege: Curious to see if this works on Windows like this :)

Cogitri: Uhh..same, but I don't have a windows maachine at my disposal

sdroege: There's Windows CI :)

Cogitri:
Hm, the tests seem to be a bit flaky with:
```
test settings::test::string_get ... 
(process:7244): GLib-GObject-CRITICAL **: 16:50:33.710: g_object_ref: assertion 'G_IS_OBJECT (object)' failed


(process:7244): GLib-GObject-CRITICAL **: 16:50:33.710: g_object_unref: assertion 'G_IS_OBJECT (object)' failed
**
GLib-GObject:ERROR:../gobject/gobject.c:4453:g_weak_ref_set: assertion failed: (weak_locations != NULL)
error: process didn't exit successfully: `/home/rasmus/Projects/gio/target/debug/deps/gio-8e55b7fc5fe6fb99` (signal: 6, SIGABRT: process abort signal)
```

build.rs
Cogitri: Oh, missed that :P

sdroege:
Can you reproduce the test flakiness yourself too or only on CI?  That looks rather problematic, probably memory unsafe somewhere.

I can take a look if it's easy to reproduce. But gdb and valgrind would be the tools of choice here.

Cogitri:
It only happens on my system from time to time while running `cargo test`

Cogitri:
Alright, seems like it happens when I:

1. run `cargo test`
2. Change something in the source file afterwards (so cargo rebuilds the test)
3. run `cargo test` again <- it happens here

Not quite sure why though

Cogitri:
Yup, it's random, just run it in a loop until it fails, takes 3-5 tries for me:

```
#!/bin/bash
cargo test
while [ $? == 0 ]; do
	cargo test
done
```

Maybe it's because the tests are run in parallel though?

sdroege:
Will take a look. It might be that the `GSettings` API is not thread-safe and we don't guard against that... but that would be rather unfortunate. I'll take a look :)

sdroege:
Also clearly not caused by your changes but something being broken before

sdroege:
Yes it seems to be caused by running multiple threads at once, awesome. Now to find out if that's a bug in gio or we must prevent usage from multiple threads at once!

src/settings.rs
```diff
@@ -0,0 +1,82 @@
...
+
+#[cfg(test)]
+mod test {
+    use super::*;
+    use std::env::set_var;
+    use std::process::Command;
+    use std::str::from_utf8;
+
+    fn set_env() {
+        let output = Command::new("glib-compile-schemas")
```
sdroege: Not that it helps, but you should wrap all code in this function in `std::sync::Once` :)

Cogitri: Oh nice, hadn't known about that, thought about using a global bool for that at first but that's way prettier, thanks :)

sdroege:
It usually explodes here for me

```
Thread 1 (Thread 0x7f57fecf5700 (LWP 31753)):
#0  0x00007f57ff504885 in _g_log_abort (breakpoint=1) at ../../../glib/gmessages.c:554
#1  0x00007f57ff505b8d in g_logv (log_domain=0x7f57ff613258 "GLib-GObject", log_level=G_LOG_LEVEL_CRITICAL, format=<optimized out>, args=args@entry=0x7f57fecf41a0) at ../../../glib/gmessages.c:1371
#2  0x00007f57ff505d5f in g_log (log_domain=log_domain@entry=0x7f57ff613258 "GLib-GObject", log_level=log_level@entry=G_LOG_LEVEL_CRITICAL, format=format@entry=0x7f57ff55557c "%s: assertion '%s' failed") at ../../../glib/gmessages.c:1413
#3  0x00007f57ff506559 in g_return_if_fail_warning (log_domain=log_domain@entry=0x7f57ff613258 "GLib-GObject", pretty_function=pretty_function@entry=0x7f57ff6167f8 <__FUNCTION__.15104> "g_object_ref", expression=expression@entry=0x7f57ff61559f "G_IS_OBJECT (object)") at ../../../glib/gmessages.c:2767
#4  0x00007f57ff5e9c4c in g_object_ref (_object=0x7f57e8002170) at ../../../gobject/gobject.c:3212
#5  0x00007f57ff5efbc5 in g_weak_ref_get (weak_ref=<optimized out>) at ../../../gobject/gobject.c:4393
#6  0x00007f57ff712df2 in g_settings_backend_invoke_closure (user_data=0x7f57e8009e80) at ../../../gio/gsettingsbackend.c:263
#7  0x00007f57ff4ffa05 in g_main_context_invoke_full (context=0x7f57f80061e0, priority=0, function=0x7f57ff712de0 <g_settings_backend_invoke_closure>, data=0x7f57e8009e80, notify=0x0) at ../../../glib/gmain.c:5843
#8  0x00007f57ff4ffa50 in g_main_context_invoke (context=<optimized out>, function=<optimized out>, data=<optimized out>) at ../../../glib/gmain.c:5788
#9  0x00007f57ff712f68 in g_settings_backend_dispatch_signal (backend=0x7f57e8009510 [GMemorySettingsBackend], function_offset=0, name=0x55cc920c0760 "/com/github/gtk-rs/test-bool", origin_tag=0x0, names=0x0) at ../../../gio/gsettingsbackend.c:330
#10 0x00007f57ff7135bb in g_settings_backend_changed (backend=<optimized out>, key=<optimized out>, origin_tag=<optimized out>) at ../../../gio/gsettingsbackend.c:379
#11 0x00007f57ff7129ce in g_memory_settings_backend_write (backend=0x7f57e8009510 [GMemorySettingsBackend], key=0x55cc920c0760 "/com/github/gtk-rs/test-bool", value=0x7f57f8009060, origin_tag=0x0) at ../../../gio/gmemorysettingsbackend.c:80
#12 0x00007f57ff713b24 in g_settings_backend_write (backend=0x7f57e8009510 [GMemorySettingsBackend], key=key@entry=0x55cc920c0760 "/com/github/gtk-rs/test-bool", value=value@entry=0x7f57f8009060, origin_tag=origin_tag@entry=0x0) at ../../../gio/gsettingsbackend.c:793
#13 0x00007f57ff718023 in g_settings_write_to_backend (settings=settings@entry=0x7f57f8007430 [GSettings], value=value@entry=0x7f57f8009060, key=<optimized out>) at ../../../gio/gsettings.c:1151
#14 0x00007f57ff719d78 in g_settings_set_value (settings=0x7f57f8007430 [GSettings], key=0x7f57f80061a0 "test-bool", value=0x7f57f8009060) at ../../../gio/gsettings.c:1591

#0  0x00007f57ff48d0df in futex_abstimed_wait (private=0, abstime=0x0, expected=2, futex_word=0x7f57f800b008) at ../sysdeps/unix/sysv/linux/futex-internal.h:172
#1  0x00007f57ff48d0df in __pthread_rwlock_wrlock_full (abstime=0x0, rwlock=0x7f57f800b000) at pthread_rwlock_common.c:803
#2  0x00007f57ff48d0df in __GI___pthread_rwlock_wrlock (rwlock=0x7f57f800b000) at pthread_rwlock_wrlock.c:27
#3  0x00007f57ff5491f5 in g_rw_lock_writer_lock (rw_lock=rw_lock@entry=0x7f57ff62c220 <weak_locations_lock>) at ../../../glib/gthread-posix.c:468
#4  0x00007f57ff5e9f02 in g_object_unref (_object=<optimized out>) at ../../../gobject/gobject.c:3279
--Type <RET> for more, q to quit, c to continue without paging--
#5  0x00007f57ff5e9f02 in g_object_unref (_object=0x7f57f8007eb0) at ../../../gobject/gobject.c:3238
```

Seems like at least the memory `GSettingsBackend` is not thread-safe... or maybe even `GWeakRef` has a thread-safety bug.

src/settings.rs
Cogitri: Pushed, thanks

sdroege:
Thread-safety bug in `GSettingsBackend`. By design it should work but the code is broken.

sdroege:
See https://gitlab.gnome.org/GNOME/glib/issues/1870 with explanation but I don't know a solution. Until that is fixed and we run a new enough version of GLib on the CI the only solution I see here for now is to disable the tests.

It's all unrelated to your code so I'd suggest to disable the tests and we just go on with life for now.

Cogitri:
Wouldn't disabling parallel run for the tests work too?

sdroege:
True, that's a better idea actually. Let's go for that then :)

sdroege:
You change the travis/appveyor configs to build only with one thread?

Cogitri:
Somehow I had hoped that I could force only one test thread in gio's crate itself so others won't hit this, but seems that that's not possible yet (unless we make our own test harness). I'll modify the configs.

Cogitri:
Unless we use something like https://crates.io/crates/serial_test

Cogitri:
Pushed fix

sdroege:
Nice

PR #227, opened 2019-07-06
New version

EPashkin:
Part of https://github.com/gtk-rs/gir-files/pull/37

cc @GuillaumeGomez, @sdroege


src/auto/functions.rs
```diff
@@ -603,6 +615,12 @@ pub fn unix_is_system_fs_type(fs_type: &str) -> bool {
 //    unsafe { TODO: call gio_sys:g_unix_mount_get_options() }
 //}
 
+//#[cfg(any(unix, feature = "dox"))]
+//#[cfg(any(feature = "v2_60", feature = "dox"))]
+//pub fn unix_mount_get_root_path(mount_entry: /*Ignored*/&mut UnixMountEntry) -> Option<GString> {
```
EPashkin: `UnixMountEntry` and `UnixMountPoint` can be generated but currently used without cfg(unix),
so build fail on windows

src/auto/output_stream.rs
```diff
@@ -128,6 +128,26 @@ pub trait OutputStreamExt: 'static {
         bytes: &glib::Bytes,
         io_priority: glib::Priority,
     ) -> Box_<dyn future::Future<Output = Result<isize, Error>> + std::marker::Unpin>;
+
+    //#[cfg(any(feature = "v2_60", feature = "dox"))]
+    //fn writev<P: IsA<Cancellable>>(&self, vectors: /*Ignored*/&[&OutputVector], cancellable: Option<&P>) -> Result<usize, Error>;
```
EPashkin: recod `GOutputVector` contains only fields and need manual implementation.

src/auto/functions.rs
sdroege: Ok but that can be fixed by configuring that or not?

sdroege:
Looks good to me otherwise. Are there any news types in the .gir that we could generate now, and if so any reason to not add them while we're at it? :)

src/auto/output_stream.rs
sdroege: Yes, and ideally with zero-cost conversion from `std::io::IoVec` :)

src/auto/functions.rs
EPashkin: Currently - not,
If I even with `cfg_condition = "unix"` on `UnixMountEntry`
in used without it `use UnixMountEntry;`
see https://github.com/gtk-rs/gio/compare/master...EPashkin:unixmount?expand=1

EPashkin:
Seems no more new types:
`GPollableReturn` used only with `GOutputVector`
`GDBusConnectionFlags` used in `GDBusConnection` and we don't generate it now.

EPashkin:
DBusConnection generates almost all: https://github.com/gtk-rs/gio/compare/master...EPashkin:dbusconnection?expand=1
but IMHO it not usable without `register_object`

sdroege:
> DBusConnection generates almost all: https://github.com/gtk-rs/gio/compare/master...EPashkin:dbusconnection?expand=1
> but IMHO it not usable without `register_object`

Let's handle that in a separate PR.

Let's get this in then once the unix stuff is solved?

EPashkin:
IMHO better solve problem `UnixMountEntry` and `UnixMountPoint` in separate PR too,
currently i has no idea how to fix this 😢

sdroege:
Ah I thought it's broken with this PR. If it needs to be enabled separately anyway, then sure let's get this in now and handle everything else separately. Thanks!

PR #229, opened 2019-07-11
Generate FileEnumerator

GuillaumeGomez:
Part of https://github.com/gtk-rs/gtk/issues/847.


src/auto/file_enumerator.rs
```diff
@@ -0,0 +1,316 @@
...
+    fn get_child(&self, info: &FileInfo) -> Option<File>;
+
+    fn get_container(&self) -> Option<File>;
+
+    fn has_pending(&self) -> bool;
+
+    fn is_closed(&self) -> bool;
+
+    #[cfg(any(feature = "v2_44", feature = "dox"))]
+    fn iterate<P: IsA<Cancellable>>(
```
sdroege: This binding is wrong and needs to be handled manually. See the docs how it's supposed to be used from C. I guess we ideally wrap this in a proper Rust iterator.

Or maybe not bind it at all, it's C convenience API and `next_file()` works better for us (and should get an `Iterator` wrapped around it).

src/auto/file_enumerator.rs
```diff
@@ -0,0 +1,316 @@
...
+        cancellable: Option<&P>,
+        callback: Q,
+    );
+
+    #[cfg(feature = "futures")]
+    fn next_files_async_future(
+        &self,
+        num_files: i32,
+        io_priority: glib::Priority,
+    ) -> Box_<dyn future::Future<Output = Result<Vec<FileInfo>, Error>> + std::marker::Unpin>;
```
sdroege: Could make a `futures::Stream` around this

src/auto/file_enumerator.rs
sdroege: Yeah I guess ignoring this function and just keeping `next_file()` and its variants would be best. But before merging, please add an

```rust
impl Iterator for FileEnumerator {
    type Item = Result<FileInfo, Error>;

    fn next(&mut self) -> Option<Result<FileInfo, Error>> {
        match self.next_file() {
            Err(err) => Some(Err(err)),
            Ok(file_info) => file_info.map(Ok),
        }
    }
}
```

With the `fallible-iterator` crate this is also good to work with.

src/auto/file_enumerator.rs
sdroege: This would be for some later time :)

GuillaumeGomez:
Updated.

src/auto/file_enumerator.rs
GuillaumeGomez: Added it.

sdroege:
:+1:

PR #230, opened 2019-07-20
Regen with nullable callbacks

GuillaumeGomez:
Requires https://github.com/gtk-rs/gir/issues/815.


sdroege:
The ones with the fds are unsafe like this. Will review in detail tomorrow
-- 
Sent from my Android device with K-9 Mail. Please excuse my brevity.

src/auto/desktop_app_info.rs
```diff
@@ -91,10 +94,27 @@ pub trait DesktopAppInfoExt: 'static {
...
+    fn launch_uris_as_manager_with_fds<P: IsA<AppLaunchContext>>(
+        &self,
+        uris: &[&str],
+        launch_context: Option<&P>,
+        spawn_flags: glib::SpawnFlags,
+        user_setup: Option<Box<dyn FnOnce() + 'static>>,
+        pid_callback: Option<&mut dyn (FnMut(&DesktopAppInfo, glib::Pid))>,
+        stdin_fd: i32,
+        stdout_fd: i32,
+        stderr_fd: i32,
```
sdroege: This should be generic over `FromRawFd` for the fds on UNIX and `FromRawHandle` on Windows. Like we do elsewhere.

Passing arbitrary numbers around as fds is unsafe.

sdroege:
Looks good otherwise

src/auto/desktop_app_info.rs
```diff
@@ -91,10 +94,14 @@ pub trait DesktopAppInfoExt: 'static {
...
-    //fn launch_uris_as_manager<P: IsA<AppLaunchContext>>(&self, uris: &[&str], launch_context: Option<&P>, spawn_flags: /*Ignored*/glib::SpawnFlags, user_setup: Option<Box_<dyn FnOnce() + 'static>>, pid_callback: /*Unimplemented*/FnMut(&DesktopAppInfo, /*Ignored*/glib::Pid), pid_callback_data: /*Unimplemented*/Option<Fundamental: Pointer>) -> Result<(), Error>;
-
-    //#[cfg(any(feature = "v2_58", feature = "dox"))]
-    //fn launch_uris_as_manager_with_fds<P: IsA<AppLaunchContext>>(&self, uris: &[&str], launch_context: Option<&P>, spawn_flags: /*Ignored*/glib::SpawnFlags, user_setup: Option<Box_<dyn FnOnce() + 'static>>, pid_callback: /*Unimplemented*/FnMut(&DesktopAppInfo, /*Ignored*/glib::Pid), pid_callback_data: /*Unimplemented*/Option<Fundamental: Pointer>, stdin_fd: i32, stdout_fd: i32, stderr_fd: i32) -> Result<(), Error>;
+    fn launch_uris_as_manager<P: IsA<AppLaunchContext>>(
+        &self,
+        uris: &[&str],
+        launch_context: Option<&P>,
+        spawn_flags: glib::SpawnFlags,
+        user_setup: Option<Box_<dyn FnOnce() + 'static>>,
```
sdroege: Not sure this is correct or safe. The function is called once per process, so if you have multiple URIs... :) OTOH this is called right after `fork()` from what I understand. Unclear if `FnOnce` is correct in this case.

src/desktop_app_info.rs
```diff
@@ -36,3 +55,177 @@ impl DesktopAppInfo {
...
+fn launch_uris_as_manager_with_fds_inner<P: IsA<AppLaunchContext>, O: IsA<DesktopAppInfo>>(
+    this: &O,
+    uris: &[&str],
+    launch_context: Option<&P>,
+    spawn_flags: glib::SpawnFlags,
+    user_setup: Option<Box_<dyn FnOnce() + 'static>>,
+    pid_callback: Option<&mut dyn (FnMut(&DesktopAppInfo, glib::Pid))>,
+    stdin_fd: FD,
+    stdout_fd: FD,
+    stderr_fd: FD,
```
sdroege: This should become a generic function with `FD: IntoRawFd`, or be marked as `unsafe`

GuillaumeGomez: Why `IntoRawFd` and not `AsRawFd`?

sdroege: You give ownership of the fds to the function or not?

GuillaumeGomez: I wouldn't, that's why I pcked `As*` instead of `Into*`.

sdroege: It doesn't matter what you would do :P What does the C function do?

GuillaumeGomez: It appears that the FDs need to be closed outside of the function: https://github.com/GNOME/gnome-shell/blob/master/src/shell-app.c#L1322

Therefore, we don't give ownership.

sdroege: Then you're right :)

src/auto/desktop_app_info.rs
sdroege: Also unclear if Rust is supposed to be fork-safe, or if the closure needs to be marked unsafe anyway

sdroege:
This is obsolete now?

GuillaumeGomez:
I guess?

GuillaumeGomez:
Wait actually it wasn't: it added new objects. Well, let's first merge your regen PR, it'll make less noise.

GuillaumeGomez:
About the failure: shouldn't we reexport all manual traits?

PR #232, opened 2019-08-10
Fix boxing in async func

EPashkin:
Part of https://github.com/gtk-rs/gir/issues/807

Include generation of UnixMountEntry and UnixMountPoint

cc @GuillaumeGomez , @sdroege


src/auto/functions.rs
```diff
@@ -26,6 +25,10 @@ use InputStream;
 use Resource;
 use ResourceLookupFlags;
 use SettingsBackend;
+#[cfg(feature = "unix")]
+use UnixMountEntry;
+#[cfg(feature = "unix")]
+use UnixMountPoint;
```
EPashkin: Seems it need `feature = "dox"` too as docs fails on https://travis-ci.org/gtk-rs/gio/jobs/570186591#L631
also usual we don't write `feature="unix"`
```
pub fn unix_mount_get_device_path(mount_entry: &mut UnixMountEntry) -> Option<std::path::PathBuf> {
```

GuillaumeGomez: An idea why the dox feature condition isn't generated?

EPashkin: Because it we missed that when futures was added:
```rust
    #[cfg(feature = "futures")]
    #[cfg(any(feature = "v2_60", feature = "dox"))]
    fn launch_uris_async_future<P: IsA<AppLaunchContext> + Clone + 'static>(
```

EPashkin:
There new problem with `UnixMountPoint`
its copy function was added later that object itself.
```
    #[cfg(any(feature = "v2_54", feature = "dox"))]
    pub fn g_unix_mount_point_copy(mount_point: *mut GUnixMountPoint) -> *mut GUnixMountPoint;
```
@GuillaumeGomez, @sdroege Can we add alternative function in sys that panics with message "copy for UnixMountPoint only implemented in Gio 2.54" ?

sdroege:
> @GuillaumeGomez, @sdroege Can we add alternative function in sys that panics with message "copy for UnixMountPoint only implemented in Gio 2.54" ?

Why not go via `g_boxed_copy()`?

EPashkin:
@sdroege You right, `g_boxed_copy` can be used instead

sdroege:
> @sdroege You right, `g_boxed_copy` can be used instead

We have some gir configuration for making use of that in autogenerated bindings btw, if that helps here

EPashkin:
@sdroege Thanks for reminder, updated code

EPashkin:
This PR is ready for review

src/auto/functions.rs
```diff
@@ -569,111 +573,193 @@ pub fn unix_is_system_fs_type(fs_type: &str) -> bool {
...
+        (ret, time_read)
+    }
+}
 
-//#[cfg(any(unix, feature = "dox"))]
-//pub fn unix_mount_free(mount_entry: /*Ignored*/&mut UnixMountEntry) {
-//    unsafe { TODO: call gio_sys:g_unix_mount_free() }
-//}
+#[cfg(any(unix, feature = "dox"))]
+pub fn unix_mount_free(mount_entry: &mut UnixMountEntry) {
```
sdroege: This is obviously unsafe :) It should be part of the `Drop` impl of `UnixMountEntry`

src/auto/functions.rs
```diff
@@ -569,111 +573,193 @@ pub fn unix_is_system_fs_type(fs_type: &str) -> bool {
...
 pub fn unix_mount_points_changed_since(time: u64) -> bool {
     unsafe { from_glib(gio_sys::g_unix_mount_points_changed_since(time)) }
 }
 
-//#[cfg(any(unix, feature = "dox"))]
-//pub fn unix_mount_points_get() -> (/*Ignored*/Vec<UnixMountPoint>, u64) {
-//    unsafe { TODO: call gio_sys:g_unix_mount_points_get() }
-//}
+#[cfg(any(unix, feature = "dox"))]
+pub fn unix_mount_points_get() -> (Vec<UnixMountPoint>, u64) {
```
sdroege: Similarly all these new functions should be part of the corresponding type's `impl`.

src/auto/functions.rs
```diff
@@ -569,111 +573,193 @@ pub fn unix_is_system_fs_type(fs_type: &str) -> bool {
...
+        ))
+    }
+}
 
-//#[cfg(any(unix, feature = "dox"))]
-//pub fn unix_mount_guess_should_display(mount_entry: /*Ignored*/&mut UnixMountEntry) -> bool {
-//    unsafe { TODO: call gio_sys:g_unix_mount_guess_should_display() }
-//}
+#[cfg(any(unix, feature = "dox"))]
+pub fn unix_mount_guess_should_display(mount_entry: &mut UnixMountEntry) -> bool {
```
sdroege: A getter probably shouldn't require a mutable reference

src/auto/functions.rs
EPashkin: `UnixMountEntry` used `g_boxed_free`, so I just ignore it

EPashkin:
Updated

src/auto/unix_mount_entry.rs
```diff
@@ -1,17 +0,0 @@
...
-use gobject_sys;
-
-glib_wrapper! {
-    #[derive(Debug, PartialEq, Eq, PartialOrd, Ord, Hash)]
-    pub struct UnixMountEntry(Boxed<gio_sys::GUnixMountEntry>);
-
-    match fn {
-        copy => |ptr| gobject_sys::g_boxed_copy(gio_sys::g_unix_mount_entry_get_type(), ptr as *mut _) as *mut gio_sys::GUnixMountEntry,
-        free => |ptr| gobject_sys::g_boxed_free(gio_sys::g_unix_mount_entry_get_type(), ptr as *mut _),
-        get_type => || gio_sys::g_unix_mount_entry_get_type(),
```
sdroege: Well this part could stay autogenerated but not too useful I guess :)

src/unix_mount_entry.rs
```diff
@@ -0,0 +1,192 @@
...
+            let mut time_read = mem::MaybeUninit::uninit();
+            let ret = FromGlibPtrContainer::from_glib_full(gio_sys::g_unix_mounts_get(
+                time_read.as_mut_ptr(),
+            ));
+            let time_read = time_read.assume_init();
+            (ret, time_read)
+        }
+    }
+
+    pub fn compare(&self, mount2: &UnixMountEntry) -> i32 {
```
sdroege: Should maybe return `std::cmp::Ordering`?

sdroege: Or rather: `impl Ord` on this type with this function :)

sdroege: Which you did

src/unix_mount_point.rs
```diff
@@ -0,0 +1,25 @@
...
+            let mut time_read = mem::MaybeUninit::uninit();
+            let ret = FromGlibPtrContainer::from_glib_full(gio_sys::g_unix_mount_points_get(
+                time_read.as_mut_ptr(),
+            ));
+            let time_read = time_read.assume_init();
+            (ret, time_read)
+        }
+    }
+
+    pub fn unix_mount_points_changed_since(time: u64) -> bool {
```
sdroege: The other one is called `is_changed_since()`. Better keep it consistent

src/unix_mount_point.rs
```diff
@@ -0,0 +1,25 @@
...
+// See the COPYRIGHT file at the top-level directory of this distribution.
+// Licensed under the MIT license, see the LICENSE file or <http://opensource.org/licenses/MIT>
+
+use glib::translate::*;
+use std::mem;
+use UnixMountPoint;
+
+impl UnixMountPoint {
+    #[cfg(any(unix, feature = "dox"))]
+    pub fn unix_mount_points_get() -> (Vec<UnixMountPoint>, u64) {
```
sdroege: The `unix_mount_points_` prefix is not really needed

src/unix_mount_entry.rs
```diff
@@ -0,0 +1,192 @@
...
+
+    pub fn get_fs_type(&self) -> Option<GString> {
+        unsafe {
+            from_glib_none(gio_sys::g_unix_mount_get_fs_type(mut_override(
+                self.to_glib_none().0,
+            )))
+        }
+    }
+
+    pub fn unix_mount_get_mount_path(&self) -> Option<std::path::PathBuf> {
```
sdroege: The `unix_mount_` prefix is not really needed

src/unix_mount_point.rs
EPashkin: Thanks, I just forgot to remove prefix in this file

EPashkin:
Updated

src/auto/unix_mount_entry.rs
EPashkin: I thought at start that I can use concrete functions but `g_unix_mount_copy` has `feature = "v2_54"` and I just forgot to move it back,
Agree that is right to generate even part of implementation because it can be simply expanded when gir evolves

EPashkin: Hm, seems I implemented it manually because generated derives PartialEq, Eq, PartialOrd, Ord and gir seems don't have way to change it.

sdroege: See https://github.com/gtk-rs/gir/issues/828#issuecomment-522209798

src/unix_mount_entry.rs
```diff
@@ -0,0 +1,192 @@
...
+            let ret = from_glib_full(gio_sys::g_unix_mount_for(
+                file_path.as_ref().to_glib_none().0,
+                time_read.as_mut_ptr(),
+            ));
+            let time_read = time_read.assume_init();
+            (ret, time_read)
+        }
+    }
+
+    pub fn get() -> (Vec<UnixMountEntry>, u64) {
```
sdroege: Maybe this should be `get_mounts()`, i.e. plural? It's not getting a single one but all of them

src/unix_mount_point.rs
```diff
@@ -0,0 +1,25 @@
...
+// See the COPYRIGHT file at the top-level directory of this distribution.
+// Licensed under the MIT license, see the LICENSE file or <http://opensource.org/licenses/MIT>
+
+use glib::translate::*;
+use std::mem;
+use UnixMountPoint;
+
+impl UnixMountPoint {
+    #[cfg(any(unix, feature = "dox"))]
+    pub fn get() -> (Vec<UnixMountPoint>, u64) {
```
sdroege: As above, maybe `get_mount_points()`?

src/unix_mount_entry.rs
sdroege: Or `get_mount_entries()`

sdroege:
Thanks, almost there I guess :)

src/unix_mount_entry.rs
EPashkin: agree

src/auto/unix_mount_entry.rs
sdroege: Want to do something about this one or merge for now?

sdroege:
Thanks, will merge once CI is happy :)

sdroege:
Are there more regens needed for this one?

PR #233, opened 2019-08-11
Update for new `Value::get` signature

fengalin:
See https://github.com/gtk-rs/glib/pull/513 & https://github.com/gtk-rs/gir/pull/825


sdroege:
:+1: Will merge once the CI passed (as far as it can pass at this point). Thanks!

fengalin:
As "good" as it can be... for now :)

PR #238, opened 2019-08-27
Add support for subclassing Input/OutputStream

sdroege:
I'll also do the `Seekable` interface in a bit. Don't merge yet, but review :)

Also see https://github.com/gtk-rs/gio/issues/160

CC @brainblasted @GuillaumeGomez @EPashkin @fkrull

### DIFF
--- a/src/read_input_stream.rs
+++ b/src/read_input_stream.rs
@@ -1,0 +1,411 @@
+// Copyright 2019, The Gtk-rs Project Developers.
+// See the COPYRIGHT file at the top-level directory of this distribution.
+// Licensed under the MIT license, see the LICENSE file or <http://opensource.org/licenses/MIT>
+
+use crate::prelude::*;
+use crate::subclass::prelude::*;
+use crate::InputStream;
+use glib;
+use glib::prelude::*;
+use glib::subclass;
+use glib::translate::*;
+
+use std::any::Any;
+use std::io::{Read, Seek};
+
+mod imp {
+    use super::*;
+    use std::cell::RefCell;
+
+    pub(super) enum Reader {
+        Read(AnyReader),
+        ReadSeek(AnyReader),
+    }
+
+    pub struct ReadInputStream {
+        pub(super) read: RefCell<Option<Reader>>,
+    }
+
+    impl ObjectSubclass for ReadInputStream {
+        const NAME: &'static str = "ReadInputStream";
+        type ParentType = InputStream;
+        type Instance = subclass::simple::InstanceStruct<Self>;
+        type Class = subclass::simple::ClassStruct<Self>;
+
+        glib_object_subclass!();
+
+        fn new() -> Self {
+            Self {
+                read: RefCell::new(None),
+            }
+        }
+
+        fn type_init(type_: &mut subclass::InitializingType<Self>) {
+            type_.add_interface::<crate::Seekable>();
+        }
+    }
+
+    impl ObjectImpl for ReadInputStream {
+        glib_object_impl!();
+    }
+
+    impl InputStreamImpl for ReadInputStream {
+        fn read(
+            &self,
+            _stream: &InputStream,
+            buffer: &mut [u8],
+            _cancellable: Option<&crate::Cancellable>,
+        ) -> Result<usize, glib::Error> {
+            let mut read = self.read.borrow_mut();
+            let read = match *read {
+                None => {
+                    return Err(glib::Error::new(
+                        crate::IOErrorEnum::Closed,
+                        "Already closed",
+                    ));
+                }
+                Some(Reader::Read(ref mut read)) => read,
+                Some(Reader::ReadSeek(ref mut read)) => read,
+            };
+
+            loop {
+                match std_error_to_gio_error(read.read(buffer)) {
+                    None => continue,
+                    Some(res) => return res,
+                }
+            }
+        }
+
+        fn close(
+            &self,
+            _stream: &InputStream,
+            _cancellable: Option<&crate::Cancellable>,
+        ) -> Result<(), glib::Error> {
+            let _ = self.read.borrow_mut().take();
+            Ok(())
+        }
+    }
+
+    impl SeekableImpl for ReadInputStream {
+        fn tell(&self, _seekable: &crate::Seekable) -> i64 {
+            // XXX: stream_position is not stable yet
+            // let mut read = self.read.borrow_mut();
+            // match *read {
+            //     Some(Reader::ReadSeek(ref mut read)) => {
+            //         read.stream_position().map(|pos| pos as i64).unwrap_or(-1)
+            //     },
+            //     _ => -1,
+            // };
+            -1
+        }
+
+        fn can_seek(&self, _seekable: &crate::Seekable) -> bool {
+            let read = self.read.borrow();
+            match *read {
+                Some(Reader::ReadSeek(_)) => true,
+                _ => false,
+            }
+        }
+
+        fn seek(
+            &self,
+            _seekable: &crate::Seekable,
+            offset: i64,
+            type_: glib::SeekType,
+            _cancellable: Option<&crate::Cancellable>,
+        ) -> Result<(), glib::Error> {
+            use std::io::SeekFrom;
+
+            let mut read = self.read.borrow_mut();
+            match *read {
+                Some(Reader::ReadSeek(ref mut read)) => {
+                    let pos = match type_ {
+                        glib::SeekType::Cur => SeekFrom::Current(offset),
+                        glib::SeekType::Set => {
+                            if offset < 0 {
+                                return Err(glib::Error::new(
+                                    crate::IOErrorEnum::InvalidArgument,
+                                    "Invalid Argument",
+                                ));
+                            } else {
+                                SeekFrom::Start(offset as u64)
+                            }
+                        }
+                        glib::SeekType::End => SeekFrom::End(offset),
+                        _ => unimplemented!(),
+                    };
+
+                    loop {
+                        match std_error_to_gio_error(read.seek(pos)) {
+                            None => continue,
+                            Some(res) => return res.map(|_| ()),
+                        }
+                    }
+                }
+                _ => Err(glib::Error::new(
+                    crate::IOErrorEnum::NotSupported,
+                    "Truncating not supported",
+                )),
+            }
+        }
+
+        fn can_truncate(&self, _seekable: &crate::Seekable) -> bool {
+            false
+        }
+
+        fn truncate(
+            &self,
+            _seekable: &crate::Seekable,
+            _offset: i64,
+            _cancellable: Option<&crate::Cancellable>,
+        ) -> Result<(), glib::Error> {
+            Err(glib::Error::new(
+                crate::IOErrorEnum::NotSupported,
+                "Truncating not supported",
+            ))
+        }
+    }
+}
+
+glib_wrapper! {
+    pub struct ReadInputStream(Object<subclass::simple::InstanceStruct<imp::ReadInputStream>, subclass::simple::ClassStruct<imp::ReadInputStream>, ReadInputStreamClass>) @extends crate::InputStream, @implements crate::Seekable;
+
+    match fn {
+        get_type => || imp::ReadInputStream::get_type().to_glib(),
+    }
+}
+
+impl ReadInputStream {
+    pub fn new<R: Read + Send + 'static>(read: R) -> ReadInputStream {
+        let obj = glib::Object::new(Self::static_type(), &[])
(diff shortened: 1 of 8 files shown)


src/subclass/input_stream.rs
```diff
@@ -0,0 +1,322 @@
...
+    fn parent_close(
+        &self,
+        stream: &InputStream,
+        cancellable: Option<&Cancellable>,
+    ) -> Result<(), Error> {
+        unsafe {
+            let data = self.get_type_data();
+            let parent_class = data.as_ref().get_parent_class() as *mut gio_sys::GInputStreamClass;
+            let mut err = ptr::null_mut();
+            if let Some(f) = (*parent_class).close_fn {
```
GuillaumeGomez: Maybe a `match` would be better in here?

GuillaumeGomez: Nevermind, it'd be an issue for `err`.

src/subclass/input_stream.rs
```diff
@@ -0,0 +1,322 @@
...
+    ptr: *mut gio_sys::GInputStream,
+    buffer: glib_sys::gpointer,
+    count: usize,
+    cancellable: *mut gio_sys::GCancellable,
+    err: *mut *mut glib_sys::GError,
+) -> isize
+where
+    T: InputStreamImpl,
+{
+    use std::slice;
```
GuillaumeGomez: Why did you put this import here?

src/subclass/input_stream.rs
```diff
@@ -0,0 +1,322 @@
...
+    let imp = instance.get_impl();
+    let wrap: InputStream = from_glib_borrow(ptr);
+
+    match imp.read(
+        &wrap,
+        slice::from_raw_parts_mut(buffer as *mut u8, count),
+        Option::<Cancellable>::from_glib_borrow(cancellable).as_ref(),
+    ) {
+        Ok(res) => {
+            use std::isize;
```
GuillaumeGomez: Same.

src/subclass/input_stream.rs
```diff
@@ -0,0 +1,322 @@
...
+    let imp = instance.get_impl();
+    let wrap: InputStream = from_glib_borrow(ptr);
+
+    match imp.skip(
+        &wrap,
+        count,
+        Option::<Cancellable>::from_glib_borrow(cancellable).as_ref(),
+    ) {
+        Ok(res) => {
+            use std::isize;
```
GuillaumeGomez: Same.

src/subclass/output_stream.rs
```diff
@@ -0,0 +1,381 @@
...
+        unsafe {
+            let data = self.get_type_data();
+            let parent_class = data.as_ref().get_parent_class() as *mut gio_sys::GOutputStreamClass;
+            let f = (*parent_class)
+                .write_fn
+                .expect("No parent class implementation for \"write\"");
+            let mut err = ptr::null_mut();
+            let res = f(
+                stream.to_glib_none().0,
+                buffer.as_ptr() as *mut u8,
```
GuillaumeGomez: Should buffer be a `&mut [u8]` then? We're modifying content of a constant slice.

src/subclass/input_stream.rs
sdroege: Because it's only used in this single function. Didn't want to pollute the global imports

src/subclass/input_stream.rs
sdroege: Same ;)

src/subclass/output_stream.rs
sdroege: We're not modifying it. It's just like `write()` in the `Write` trait. The bytes are written out to somewhere else. It's missing the `const` in C, that's the only problem here.

sdroege:
Also support for `gio::Seekable` is there now. Next: wrappers for `Read`/`Write`/`Seekable` :)

Please review again! I consider the remaining comments above resolved, unless someone disagrees.

src/subclass/output_stream.rs
EPashkin: Maybe use `mut_override`

EPashkin:
@sdroege Thanks LGFM

src/subclass/input_stream.rs
```diff
@@ -295,6 +301,58 @@ mod tests {
         }
     }
 
+    impl SeekableImpl for SimpleInputStream {
+        fn tell(&self, _seekable: &Seekable) -> i64 {
+            *self.pos.borrow() as i64
+        }
+
+        fn can_seek(&self, _seekable: &Seekable) -> bool {
+            true
```
GuillaumeGomez: Isn't there supposed to be a limit to how much you can seek? Or is it just a global "is this object seekable"?

src/subclass/input_stream.rs
```diff
@@ -295,6 +301,58 @@ mod tests {
...
+        ) -> Result<(), glib::Error> {
+            let mut pos = self.pos.borrow_mut();
+            match type_ {
+                glib::SeekType::Set => {
+                    *pos = offset as usize;
+                    Ok(())
+                }
+                glib::SeekType::Cur => {
+                    if offset < 0 {
+                        *pos -= (-offset) as usize;
```
GuillaumeGomez: Shouldn't we check if offset is superior than pos in here?

src/subclass/input_stream.rs
sdroege: Just globally at this very moment

src/subclass/input_stream.rs
sdroege: It's in the test, and this will panic if it fails (overflows/etc panic in debug/test builds)

src/subclass/input_stream.rs
GuillaumeGomez: Well, not sure imports are supposed to work this way but I don't really mind...

src/subclass/input_stream.rs
GuillaumeGomez: You know I'm not a big fan of panic. :p

Shouldn't we just use `max(pos, (-offset) as usize)`?

sdroege: Well, this is a test. It's supposed to panic if things fail, just like all the `assert!` later in the test. Do you want me to make this more explicit by adding assertions here?

In a real implementation you would return a `glib::Error` here.

sdroege:
The last thing I'm going to add (probably tomorrow) is a `WriteOutputStream` (basically the inverse of the `ReadInputStream`), and then subclassing support for `gio::IOStream`. Then this here should be ready to be merged.

src/subclass/input_stream.rs
GuillaumeGomez: Oh crap! I thought it was still "normal" code and not test. Nevermind and sorry for the noise...

src/read_input_stream.rs
```diff
@@ -0,0 +1,307 @@
...
+                            }
+                        }
+                        glib::SeekType::End => SeekFrom::End(offset),
+                        _ => unimplemented!(),
+                    };
+
+                    loop {
+                        match std_error_to_gio_error(read.seek(pos)) {
+                            None => continue,
+                            Some(res) => return res.map(|_| ()),
```
sdroege: Here we need to also check the `Ok` case: the actual position seeked to might be different than the requested position.

sdroege: Actually no: if it succeeds then it is at the requested position. The return position is only there to get an absolute value for the new position in case a seek relative to the end or current position was done.

src/subclass/io_stream.rs
```diff
@@ -0,0 +1,162 @@
...
+where
+    T: IOStreamImpl,
+{
+    let instance = &*(ptr as *mut T::Instance);
+    let imp = instance.get_impl();
+    let wrap: IOStream = from_glib_borrow(ptr);
+
+    let ret = imp.get_input_stream(&wrap);
+
+    // FIXME: We somehow need to ensure the stream stays alive
```
sdroege: This needs some thinking: we need to ensure that the returned stream stays alive as long as the instance. Could go via `GObject` qdata but that would easily cause circular references and thus memory leaks.

Needs some thinking.

src/write_output_stream.rs
```diff
@@ -0,0 +1,296 @@
...
+            .expect("Created write input stream is of wrong type");
+
+        let imp = imp::WriteOutputStream::from_instance(&obj);
+        *imp.write.borrow_mut() = Some(imp::Writer::WriteSeek(Box::new(write)));
+
+        obj
+    }
+
+    // TODO: Need to implement something to get back the inner
+    //pub fn take_inner(&self) -> Box<dyn std::any::Any + 'static> {
```
sdroege: Here and in the `ReadInputStream` we need something to get back the inner part. Should be done somewhat similar to how we do it in cairo already, which would also allow handling of panics in a nicer way.

src/subclass/io_stream.rs
sdroege: qdata would actually be fine for it in this case

EPashkin: Good to hear that

sdroege: Unfortunately means that this specific one can't be autogenerated later :)

sdroege:
Should be good for the final reviews now, functionality and tests are all implemented and passing.

src/read_input_stream.rs
```diff
@@ -0,0 +1,411 @@
...
+        let _ = self.close(crate::NONE_CANCELLABLE);
+
+        match ret {
+            AnyOrPanic::Any(r) => r,
+            AnyOrPanic::Panic(p) => std::panic::resume_unwind(p),
+        }
+    }
+}
+
+enum AnyOrPanic {
```
EPashkin: Duplicate?

sdroege: You mean the two versions in `read_input_stream.rs` and `write_output_stream.rs`? Yes, it seemed cleaner to duplicate those 4 lines than exporting them from the module. At later times we might also need to have them look different.

EPashkin: Ok, it private anyway

EPashkin:
@sdroege LGFM, thanks for big work

sdroege:
> @sdroege LGFM, thanks for big work

Thanks for reviewing. I just added the `Fixes #160` to the last two commit messages. Will merge once @GuillaumeGomez is also happy with it.

src/read_input_stream.rs
```diff
@@ -0,0 +1,411 @@
...
+            _cancellable: Option<&crate::Cancellable>,
+        ) -> Result<(), glib::Error> {
+            let _ = self.read.borrow_mut().take();
+            Ok(())
+        }
+    }
+
+    impl SeekableImpl for ReadInputStream {
+        fn tell(&self, _seekable: &crate::Seekable) -> i64 {
+            // Not stream_position is not stable yet
```
GuillaumeGomez: "Not stream_position is not stable yet"?

src/subclass/input_stream.rs
```diff
@@ -0,0 +1,396 @@
...
+    let wrap: InputStream = from_glib_borrow(ptr);
+
+    match imp.read(
+        &wrap,
+        slice::from_raw_parts_mut(buffer as *mut u8, count),
+        Option::<Cancellable>::from_glib_borrow(cancellable).as_ref(),
+    ) {
+        Ok(res) => {
+            use std::isize;
+            assert!(res < isize::MAX as usize);
```
GuillaumeGomez: Not a big fan of `assert` calls. :-/

sdroege: If we don't assert here, then C will assert one step further :) It's broken API at the C level.

And more important, this case can't ever happen at all: before calling `stream_read()`, C will check that the `count` is smaller than `isize::MAX`. We will never ever get called with a bigger value than that, so if the subclass reads more than there is memory available then this is a serious bug in the subclass that we can't do anything other than panicking.

You prefer segfaults over panics? :)

sdroege: But thanks for reminding me to add some more assertions here to ensure things don't fall apart.

GuillaumeGomez: No, I prefer `Result`s over panics. :3

sdroege: We have this discussion over and over again. There's a difference between recoverable/handleable errors (-> `Result`) and fatal errors (-> panics). This is the second category :)

GuillaumeGomez: I know but I need to make my point. \*anti-panic instinct\*

sdroege: Added some more assertions to make things more explicit. All these parts are also checked on the C side and causing assertions there, having them on our side only makes these constraints more clear.

GuillaumeGomez:
Thanks!

PR #241, opened 2019-09-17
Remove ignoring not generated property

EPashkin:
Part of https://github.com/gtk-rs/gtk/issues/892

I hope that CI will pass without updating submodules

cc @GuillaumeGomez, @sdroege

### DIFF
--- a/Gir.toml
+++ b/Gir.toml
@@ -47,6 +47,7 @@ generate = [
     "Gio.FileOutputStream",
     "Gio.FileQueryInfoFlags",
     "Gio.FileType",
+    "Gio.FilterInputStream",
     "Gio.Icon",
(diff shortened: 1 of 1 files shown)


sdroege:
What's the point of this change, and why doesn't this generate the `base-stream` property as a read-only (or rather: construct-only) property? :)

EPashkin:
It already has getter so ignoring was NOP
https://github.com/gtk-rs/gio/blob/master/src/auto/filter_input_stream.rs#L41
https://github.com/gtk-rs/gio/blob/master/src/auto/filter_output_stream.rs#L41
But with old version gir don't add "base_stream" to child builders

sdroege:
Ah I see. Makes sense! Will merge once CI is happy.

PR #242, opened 2019-09-24
IsA<Cancellable> generic instead of concrete Option<&Cancellable>

### LINKED ISSUE #234: Manual bindings for functions with Cancellable args are inconsistent
For functions which take a `Cancellable`, auto-generated code looks like this:

```rust
fn foo<P: IsA<Cancellable>>(cancellable: Option<&P>) { ... }
```

But in some places in the manually-written bindings, there is this instead:

```rust
fn bar(cancellable: Option<&Cancellable>) { ... }
```

The following are the inconsistent ones I could find in gio:

```
PollableOutputStreamExtManual
  create_source
  create_source_future
  create_source_stream

InputStreamExtManual
  read
  read_all
  read_all_async
  read_async
  
Subprocess
  communicate_utf8_async
  

impl<F, O, T: 'static, E: 'static> GioFuture<F, O, T, E>
where
    O: Clone + 'static,
    F: FnOnce(&O, oneshot::Sender<Result<T, E>>) -> Cancellable + 'static,
    not sure if it makes sense to let the operation return an IsA<Cancellable> here

    same for impl<F, O, T, E> Future for GioFuture<F, O, T, E>

SocketListenerExtManual
  accept_socket_async
  
SocketExtManual
  receive
  receive_from
  receive_with_blocking
  send
  send_to
  send_with_blocking
  create_source
  create_source_future
  create_source_stream

PollableInputStreamExtManual
  create_source
  create_source_future
  create_source_stream
  read_nonblocking
  
OutputStreamExtManual
  write_async
  write_all
  write_all_async
  
FileExtManual
  replace_contents_async
```

federicomenaquintero:
Fixes #234

### DIFF
--- a/src/file.rs
+++ b/src/file.rs
@@ -21,13 +21,14 @@ pub trait FileExtManual: Sized {
     fn replace_contents_async<
         B: AsRef<[u8]> + Send + 'static,
         R: FnOnce(Result<(B, glib::GString), (B, Error)>) + Send + 'static,
+        C: IsA<Cancellable>,
     >(
         &self,
         contents: B,
         etag: Option<&str>,
         make_backup: bool,
         flags: FileCreateFlags,
-        cancellable: Option<&Cancellable>,
+        cancellable: Option<&C>,
         callback: R,
     );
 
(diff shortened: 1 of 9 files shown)


federicomenaquintero:
I just need to know if I should do the same on the relatively new subclass/*stream.rs :)

EPashkin:
IMHO `InputStreamImplExt` etc. don't need `IsA<Cancelable>` to minimize code size

sdroege:
> I just need to know if I should do the same on the relatively new subclass/*stream.rs :)

See https://github.com/gtk-rs/gio/issues/234#issuecomment-534439952. Short answer is "no" :)

sdroege:
@federicomenaquintero the build is failing in all configurations it seems :)

federicomenaquintero:
Sorry, I forgot the tests!  Fixed now.

sdroege:
> Sorry, I forgot the tests! Fixed now.

Some more are fixed now, but I think you forgot at least the futures code :) See https://travis-ci.org/gtk-rs/gio/jobs/589194656

Try running tests with `--features=v2_60,futures`

federicomenaquintero:
Uuuh... is that a cargo bug in the MacOS builds?

sdroege:
> Uuuh... is that a cargo bug in the MacOS builds?

```
thread 'main' panicked at 'index out of bounds: the len is 2 but the index is 2', src/tools/cargo/src/cargo/util/cpu.rs:152:29

note: run with `RUST_BACKTRACE=1` environment variable to display a backtrace.
```

Is clearly a bug in cargo, yes. Feel like reporting it? :D

federicomenaquintero:
Thanks!  I'll try a run in my private repo with RUST_BACKTRACE=1 in the Travis environment to see if that gets useful info out of cargo panicking...

PR #249, opened 2019-10-28
Don't re-export glib types

sdroege:
Depends on: https://github.com/gtk-rs/gir/pull/866


sdroege:
CC @EPashkin @GuillaumeGomez

src/subclass/mod.rs
```diff
@@ -14,5 +14,4 @@ pub mod prelude {
     pub use super::io_stream::IOStreamImpl;
     pub use super::output_stream::OutputStreamImpl;
     pub use super::seekable::SeekableImpl;
-    pub use glib::subclass::prelude::*;
```
sdroege: Now this one here needs some discussions. Should we always re-export preludes of dependencies from our own preludes, or not?

It seems useful to do but then we should also do that for `gio::prelude`. Opinions?

GuillaumeGomez: I always wondered about this... Is it better to reexport dependencies public interfaces allowing users to not have to add dependencies themselves or should we keep things as they are? I really don't know...

sdroege: Personally I think it's nicer to re-export public dependencies from the preludes. That way you can at least use all the base class' APIs directly. It makes peoples life easier and less unexpected: I'd expect to be able to use GObject API directly on a `gtk::Widget` at least.

GuillaumeGomez: Well then, since I'm neutral, it sounds like we're more interested into reexporting. What do you think @EPashkin ?

EPashkin: As we begin to remove reexport glib Error, Inhibit etc. IMHO better continue this way.

While don't add explicit dependency is simple, it can cause misunderstanding (these type is same?) etc.
So I prefer don't reexport types if possible.

sdroege: Do you have an example how it can cause misunderstanding in the case of the preludes where you never use the types directly, but only via traits?

My main argument here why this is different is that you never use the types directly, and from other languages you'd expect to be able to use the base class API of every object you have without having to import anything new and additional.

EPashkin: Example of misunderstanding with types: https://github.com/gtk-rs/gtk/issues/897
I seems was too focused to this case and missed main point: it not types but traits.
I think that reexporting traits is good idea.

sdroege: Ok, then I'll do that consistently and update all the crates one after another in the next days :) Thanks!

sdroege:
Everybody happy with this PR before I prepare the other crates? :)

EPashkin:
@sdroege Thanks, 👍

sdroege:
@GuillaumeGomez ?

GuillaumeGomez:
:+1:

PR #250, opened 2019-11-02
Regen and prelude fixes

sdroege:


sdroege:
CC @GuillaumeGomez @EPashkin

EPashkin:
Seems examples need fix first

EPashkin:
@sdroege Thanks, 👍 after CI

GuillaumeGomez:
About the failure: shouldn't we reexport all manual traits?

GuillaumeGomez:
It makes no sense to me to be able to have `gio::Trait` but not `gio::TraitManual`...

sdroege:
> It makes no sense to me to be able to have `gio::Trait` but not `gio::TraitManual`...

Yes. We don't do that? I'll check tomorrow, I might be confused :)

sdroege:
Oh I see. Yes the way how the prelude was re-exported was actually good and we should keep that. Sorry for the confusion :)

So only the regeneration here now. Will merge once green

PR #255, opened 2019-11-11
Allow borrowing the input buffer from the read/write futures

sdroege:
This only works with async/await now instead of the combinators but
allows for much nicer usage:

let mut buf = [0u8; 1024];
let len = istream.read(&mut buf[..]).await;
println!("{:?}", &buf[0..len]);

vs previously requiring owned buffers to be passed around.


EPashkin:
@sdroege Thanks

sdroege:
See https://github.com/gtk-rs/examples/pull/268 for what this allows.

sdroege:
And added some more cleanup here in addition

sdroege:
Yeah unfortunately this is unsafe :) Cancellation of gio futures is not instantaneous, so the buffer might still be used after the future is dropped... i.e. the memory might be used after it was freed.

Let's forget about this then :) I have a different approach coming up soon.

PR #256, opened 2019-11-11
Some updates to the I/O streams

sdroege:
More to come


sdroege:
@GuillaumeGomez Please review :) I'll send a PR with an example for the examples later.

PR #270, opened 2020-01-04
Add an AsyncRead+AsyncWrite adapter for IOStream

sdroege:
In many places a combination of both is required so let's make this a
    bit simpler for users.

----

That's basically most of the boilerplate in the gio_async_tls.rs example. I'll update that once this one here is merged.

### DIFF
--- a/src/pollable_output_stream.rs
+++ b/src/pollable_output_stream.rs
@@ -155,11 +155,11 @@ pub struct OutputStreamAsyncWrite<T: IsA<PollableOutputStream>>(
 );
 
 impl<T: IsA<PollableOutputStream>> OutputStreamAsyncWrite<T> {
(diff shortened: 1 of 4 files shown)


src/pollable_output_stream.rs
```diff
@@ -155,11 +155,11 @@ pub struct OutputStreamAsyncWrite<T: IsA<PollableOutputStream>>(
 );
 
 impl<T: IsA<PollableOutputStream>> OutputStreamAsyncWrite<T> {
-    pub fn into_input_stream(self) -> T {
+    pub fn into_output_stream(self) -> T {
```
EPashkin: Good catch

EPashkin:
@sdroege Thanks, 👍

GuillaumeGomez:
Thanks!

PR #279, opened 2020-01-30
Restore and deprecate wrong named functions

EPashkin:
cc @sdroege


sdroege:
Yes that's what I meant, but only for the bugfix release. For master it's ok to not have it :)

EPashkin:
For master not adding is ok only if we plan to remove deprecated in next major,
maybe let it stay for one-two mayor ? cc @GuillaumeGomez too.
Also I copied name to wrong place fixed.

sdroege:
> For master not adding is ok only if we plan to remove deprecated in next major,

That's the plan. It's simply wrong and trivial to fix for people when updating. Let's not add it back in master, but it would probably be nice to have the wrong name deprecated in the bugfix release and the correct name available.

EPashkin:
Ok, lets close it then

sdroege:
Instead of closing, rebase it on the 0.9 branch :)

GuillaumeGomez:
Let's just change things in master, yes.

PR #281, opened 2020-02-18
Fix/extend DataInputStream code

sdroege:
See individual commits.


sdroege:
CC @EPashkin @GuillaumeGomez

GuillaumeGomez:
Sorry, I already looked at it but I can't provide a full review and be sure it's working. :-/ Do you have some precise points you want to highlight so I can agree that they are working as expected?

EPashkin:
@sdroege Thanks, LGFM
Can you add `manual_traits = ["DataInputStreamExtManual"]` to Gir.toml too?

GuillaumeGomez:
@EPashkin Great catch! We should definitely add a check to be sure we don't forget it.

I'll open an issue about it.

sdroege:
>  Can you add `manual_traits = ["DataInputStreamExtManual"]` to Gir.toml too?

Done. Thanks for the reviews, I'll merge then after the CI is happy again :)

PR #284, opened 2020-03-12
Credentials::get_unix_user returns user id

EPashkin:
Part of https://github.com/gtk-rs/gir/pull/891#issuecomment-598383431

cc @sdroege


sdroege:
:+1:

src/auto/credentials.rs
```diff
@@ -40,12 +40,12 @@ impl Credentials {
     }
 
     #[cfg(any(unix, feature = "dox"))]
-    pub fn get_unix_user(&self) -> Result<(), glib::Error> {
+    pub fn get_unix_user(&self) -> Result<u32, glib::Error> {
```
GuillaumeGomez: Just thinking about it but shouldn't we create a `UserId` type or something?

EPashkin: Even if we add it, this type need be freely convertible from/to u32 anyway.
also we support "type" change only for return now.

GuillaumeGomez: True, and the fact that all unix types are 32 bits integers make it worse... But at least it'd be more easy to prevent having someone freely passing a pid as a user_id (even though they totally can).

sdroege: In `std` it's also just using a `u32`, see e.g. [std::os::unix::process::CommandExt::uid()](https://doc.rust-lang.org/std/os/unix/process/trait.CommandExt.html#tymethod.uid)

I'd keep it as it is for now.

src/subclass/mod.rs
```diff
@@ -9,10 +9,10 @@ pub mod output_stream;
 pub mod seekable;
 
 pub mod prelude {
-    pub use super::application::{ApplicationImpl, ArgumentList};
-    pub use super::input_stream::InputStreamImpl;
-    pub use super::io_stream::IOStreamImpl;
-    pub use super::output_stream::OutputStreamImpl;
+    pub use super::application::{ApplicationImpl, ApplicationImplExt, ArgumentList};
```
sdroege: I wonder why `ArgumentList` is here. Consistent with other things it should be directly in `mod.rs` and not in the prelude as it's not a trait.

src/auto/credentials.rs
sdroege: @GuillaumeGomez ok for you?

GuillaumeGomez: Well, if the std does it (badly) this way, then it's better to align ourselves with it. So yes, it's ok for me.

src/subclass/mod.rs
EPashkin: `pub use application::ArgumentList;` outside `mod prelude` or only remove it from prelude ?
There are no other structs in subclass/*

sdroege: I'd move it from the prelude to the subclass module

EPashkin: Same question about glib, it has `Boxed`. `Property`, `InstanceStruct`, `ClassStruct`, `InitializingType`, `TypeData`, `SignalInvocationHint`, `SignalClassHandlerToken`.

EPashkin: I not sure that import them by full path is good idea.
IMHO better reexport them and prelude, hide submodules and use as `glib::subclass::Boxed` etc.

sdroege: Sounds good to me

EPashkin: Done

EPashkin:
examples needed fix beforehand

sdroege:
> @EPashkin pushed 0 commits.

How did you do that? :D

EPashkin:
I removed last commit with removal of "unused attribute" (wrong detection in stable/beta/nightly) in
```
    #[test]
    #[should_panic]
    #[serial]
    fn wrong_type() {
```

EPashkin:
Ok, now warning removed and CI passed

PR #286, opened 2020-04-05
Update for new from_glib_borrow signature

sdroege:
See https://github.com/gtk-rs/glib/pull/605


EPashkin:
👍 Builds for me locally

PR #289, opened 2020-04-13
Add callback check

GuillaumeGomez:
This PR is meant to fail until we put back the old function pointer cast system. It'll be used to enforce that at least this cast will be correct.

cc @EPashkin @sdroege


GuillaumeGomez:
Updated.

EPashkin:
👍 test works

sdroege:
@GuillaumeGomez Maybe merge this the regenerate PR?

GuillaumeGomez:
I prefer not to. I'll certainly make a PR in glib with the same purpose and then use one of them in gir CI.

PR #292, opened 2020-04-25
SubProcess::communicate_utf8_async() is nullable for the returned std…

sdroege:
…in/stderr strings

----

Same bug in the autogenerated code with the `glib::Bytes`... @GuillaumeGomez didn't you fix that in the code generator at some point? It's annotated correctly.


EPashkin:
👍

sdroege:
Nobody was reading my comment? :P This is not complete yet.

GuillaumeGomez:
I don't remember doing anything on async functions except adding the init macro call? And the current code still looks good to me. :)

EPashkin:
I read but I think that you need separate fix for it.
And PR has no `WIP:` 😉

sdroege:
@GuillaumeGomez I was thinking of https://github.com/gtk-rs/gir/pull/742 . So that doesn't seem to fully work :) Any ideas what is different here?

GuillaumeGomez:
I think callbacks for non-async functions are handled differently. Please open an issue on gir (and add it into the release issue too please :) ).

sdroege:
I'll take a look in a bit. Why are all these handled differently :)

GuillaumeGomez:
The conversions are different iirc. However I don't know what started this split. Maybe it's time to put it back?

sdroege:
>  I don't know what started this split

I think you did when adding support for generating the (non-async) callbacks :)

GuillaumeGomez:
That'd be surprising because we had the callbacks generated before we added support for the async functions. But maybe I'm wrong, not sure... In any case, it'd be nice to see how similar both codes are and check if they can be merged.

sdroege:
It was the other way around, we first had async functions :)

GuillaumeGomez:
Then I completely missed it, my bad...

sdroege:
Looking in any case, shouldn't be too hard to add this. Merging the two code paths is more tricky, we should do that together with moving to `quote` for generating the code.

sdroege:
Regenerated with updated gir, so let's get this in once the CI is happy

PR #294, opened 2020-05-02
Support various things which needed glib::VariantDict

### LINKED ISSUE #291: ApplicationExt::connect_handle_local_options() is missing
For the [`handle-local-options` signal](https://developer.gnome.org/gio/stable/GApplication.html#GApplication-handle-local-options).

kinnison:
In https://github.com/gtk-rs/glib/pull/634 we added glib::VariantDict so this PR turns on support for it.

This ought to fix #291 which is on the current release roadmap (https://github.com/gtk-rs/release/issues/115)

### DIFF
--- a/src/subclass/application.rs
+++ b/src/subclass/application.rs
@@ -9,10 +9,12 @@ use glib::translate::*;
 
 use glib::subclass::prelude::*;
 
+use glib::VariantDict;
+
 use Application;
 use ApplicationClass;
 
-use libc::{c_char, c_void};
+use libc::{c_char, c_int, c_void};
 use std::convert;
 use std::ffi::OsString;
 use std::fmt;
@@ -123,6 +125,10 @@ pub trait ApplicationImpl: ApplicationImplExt + 'static {
     fn startup(&self, application: &Application) {
         self.parent_startup(application)
     }
+
+    fn handle_local_options(&self, application: &Application, options: &VariantDict) -> i32 {
+        self.parent_handle_local_options(application, options)
+    }
 }
 
 pub trait ApplicationImplExt {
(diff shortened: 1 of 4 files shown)


sdroege:
Do you feel like also adding the `handle_local_options` virtual function to `src/subclass/application.rs`?

sdroege:
Looks good to me otherwise :) Thanks!

src/subclass/application.rs
```diff
@@ -281,6 +288,17 @@ impl<T: ApplicationImpl + ObjectImpl> ApplicationImplExt for T {
...
         }
     }
+
+    fn parent_handle_local_options(&self, application: &Application, options: &VariantDict) -> i32 {
+        unsafe {
+            let data = self.get_type_data();
+            let parent_class = data.as_ref().get_parent_class() as *mut gio_sys::GApplicationClass;
+            let f = (*parent_class)
+                .handle_local_options
+                .expect("No parent class implementation for \"startup\"");
```
sdroege, line 298:
```suggestion
                .expect("No parent class implementation for \"handle_local_options\"");
```

src/subclass/application.rs
```diff
@@ -434,6 +453,20 @@ where
...
+    options: *mut glib_sys::GVariantDict,
+) -> c_int
+where
+    T: ApplicationImpl,
+{
+    let instance = &*(ptr as *mut T::Instance);
+    let imp = instance.get_impl();
+    let wrap: Borrowed<Application> = from_glib_borrow(ptr);
+
+    imp.handle_local_options(&wrap, &from_glib_none(options))
```
sdroege, line 467:
```suggestion
    imp.handle_local_options(&wrap, &from_glib_borrow(options))
```

src/subclass/application.rs
sdroege, line 298: Instead of exploding, return `-1` if there's none. That's what should be the normal behavior from reading the C code.

kinnison:
I can't believe I missed those, thanks for the -1 suggestion, all done.

src/subclass/application.rs
sdroege: @kinnison This one is still pending :)

kinnison: Erm, is it?  Check the code tab again.

sdroege: Indeed, github UI confused me again!

PR #295, opened 2020-05-03
Clippy-clean plus CI

kinnison:
With https://github.com/gtk-rs/lgpl-docs/pull/73 landed, this ought to bring `gio` to the point of clippy clean with CI to enforce that.


EPashkin:
@kinnison Thanks 👍

PR #298, opened 2020-05-19
input_stream: add AsyncBufRead adapter

gdesmott:
@sdroege : finally found a solution by adding a `Transitioning` state. We should wait for me to implement `AsyncBufRead` before merging but I'd be interested about your feedback already, especially if you know of a better pattern to handle state changes.

### DIFF
--- a/src/input_stream.rs
+++ b/src/input_stream.rs
@@ -3,6 +3,9 @@
 // Licensed under the MIT license, see the LICENSE file or <http://opensource.org/licenses/MIT>
 
 use error::to_std_io_result;
+use futures_core::task::{Context, Poll};
+use futures_io::{AsyncBufRead, AsyncRead};
+use futures_util::future::FutureExt;
 use gio_sys;
 use glib::object::IsA;
 use glib::translate::*;
@@ -80,6 +83,13 @@ pub trait InputStreamExtManual: Sized {
         Self: IsA<InputStream>,
     {
         InputStreamRead(self)
+    }
+
+    fn into_async_buf_read(self, buffer_size: usize) -> InputStreamAsyncBufRead<Self>
+    where
+        Self: IsA<InputStream>,
+    {
+        InputStreamAsyncBufRead::new(self, buffer_size)
     }
 }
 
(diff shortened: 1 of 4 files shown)


src/input_stream.rs
```diff
@@ -341,6 +351,124 @@ impl<T: IsA<InputStream> + IsA<Seekable>> io::Seek for InputStreamRead<T> {
...
+        cx: &mut Context,
+        buf: &mut [u8],
+    ) -> Poll<io::Result<usize>> {
+        let reader = self.get_mut();
+        let pending = reader.set_reading();
+
+        match pending.poll_unpin(cx) {
+            Poll::Ready(Ok((out_buf, res))) => {
+                buf[..res].copy_from_slice(&out_buf[..res]);
+                reader.set_waiting(out_buf);
```
sdroege, line 458: If the `read()` was less than the buffer size you would lose the remaining bytes here. You need another state "have data". As long as you have > 0 bytes left, you would then directly return them from `poll_read()` before doing the next read.

I guess this also means that your test is missing some cases :)

src/input_stream.rs
```diff
@@ -341,6 +351,124 @@ impl<T: IsA<InputStream> + IsA<Seekable>> io::Seek for InputStreamRead<T> {
...
+        self: Pin<&mut Self>,
+        cx: &mut Context,
+        buf: &mut [u8],
+    ) -> Poll<io::Result<usize>> {
+        let reader = self.get_mut();
+        let pending = reader.set_reading();
+
+        match pending.poll_unpin(cx) {
+            Poll::Ready(Ok((out_buf, res))) => {
+                buf[..res].copy_from_slice(&out_buf[..res]);
```
sdroege, line 457: It might be that `res` > `buf.len()`

src/input_stream.rs
```diff
@@ -341,6 +351,124 @@ impl<T: IsA<InputStream> + IsA<Seekable>> io::Seek for InputStreamRead<T> {
...
+                reader.set_waiting(out_buf);
+                Poll::Ready(Ok(res))
+            }
+            Poll::Ready(Err((_, err))) => {
+                let kind = err.kind::<crate::IOErrorEnum>().unwrap();
```
sdroege, line 462: Maybe you want to set your state to something reasonable here then? Some kind of error state that immediately errors in the future?

sdroege:
> I'd be interested about your feedback already, especially if you know of a better pattern to handle state changes

I think keeping some "must not be observed" state is a good pattern here. Whenever this state would happen on a external call into your code that means that it panicked before during a transition.

You could also think about having some local struct that tries to recover the situation in its `Drop` impl so that future calls can work correctly. I saw that done in various projects.

src/input_stream.rs
gdesmott: I took a mental note to fix this and then, of course, totally forgot about it. Done.

gdesmott:
> You could also think about having some local struct that tries to recover the situation in its Drop impl so that future calls can work correctly. I saw that done in various projects.

I don't understand on which struct this `Drop` would be implemented.
Or should I fail and set the adapter state to `Failed` instead of panicing?

sdroege:
> I don't understand on which struct this `Drop` would be implemented.

You'd define a struct inside your function that impls `Drop`. You `mem::forget()` or otherwise "defuse" it at the end of your function. If it's actually dropped you know that a panic has happened and can fix up your state.

>  Or should I fail and set the adapter state to `Failed` instead of panicing?

The only case how that can happen is if a panic has already happened, so panicking seems like the right approach.

gdesmott:
I see, smart! Is it really worth it here? All my code is safe so is there chances for it to panic in the gio reading code?

sdroege:
> I see, smart! Is it really worth it here? All my code is safe so is there chances for it to panic in the gio reading code?

I don't see the possibility of panics there unless things are completely broken. I think it's fine as is.

gdesmott:
I finally managed to implement both traits. Tests are working fine when run individually but in batch I get this kind of panic:

```
thread 'async_buf_read' panicked at 'Value accessed from different thread than where it was created', /home/cassidy/.cargo/git/checkouts/glib-928cf7b282977403/fe5f2aa/src/lib.rs:255:13
stack backtrace:
   0: backtrace::backtrace::libunwind::trace
             at /cargo/registry/src/github.com-1ecc6299db9ec823/backtrace-0.3.44/src/backtrace/libunwind.rs:86
   1: backtrace::backtrace::trace_unsynchronized
             at /cargo/registry/src/github.com-1ecc6299db9ec823/backtrace-0.3.44/src/backtrace/mod.rs:66
   2: std::sys_common::backtrace::_print_fmt
             at src/libstd/sys_common/backtrace.rs:78
   3: <std::sys_common::backtrace::_print::DisplayBacktrace as core::fmt::Display>::fmt
             at src/libstd/sys_common/backtrace.rs:59
   4: core::fmt::write
             at src/libcore/fmt/mod.rs:1063
   5: std::io::Write::write_fmt
             at src/libstd/io/mod.rs:1426
   6: std::sys_common::backtrace::_print
             at src/libstd/sys_common/backtrace.rs:62
   7: std::sys_common::backtrace::print
             at src/libstd/sys_common/backtrace.rs:49
   8: std::panicking::default_hook::{{closure}}
             at src/libstd/panicking.rs:204
   9: std::panicking::default_hook
             at src/libstd/panicking.rs:224
  10: std::panicking::rust_panic_with_hook
             at src/libstd/panicking.rs:470
  11: std::panicking::begin_panic
             at /rustc/8d69840ab92ea7f4d323420088dd8c9775f180cd/src/libstd/panicking.rs:397
  12: glib::ThreadGuard<T>::get_mut
             at /home/cassidy/.cargo/git/checkouts/glib-928cf7b282977403/fe5f2aa/src/lib.rs:255
  13: <glib::main_context_futures::FutureWrapper as core::future::future::Future>::poll
             at /home/cassidy/.cargo/git/checkouts/glib-928cf7b282977403/fe5f2aa/src/main_context_futures.rs:35
  14: futures_util::future::future::FutureExt::poll_unpin
             at /home/cassidy/.cargo/registry/src/github.com-1ecc6299db9ec823/futures-util-0.3.5/src/future/future/mod.rs:554
  15: glib::main_context_futures::TaskSource::poll::{{closure}}
             at /home/cassidy/.cargo/git/checkouts/glib-928cf7b282977403/fe5f2aa/src/main_context_futures.rs:211
  16: glib::main_context::<impl glib::auto::main_context::MainContext>::with_thread_default
             at /home/cassidy/.cargo/git/checkouts/glib-928cf7b282977403/fe5f2aa/src/main_context.rs:124
  17: glib::main_context_futures::TaskSource::poll
             at /home/cassidy/.cargo/git/checkouts/glib-928cf7b282977403/fe5f2aa/src/main_context_futures.rs:205
  18: glib::main_context_futures::TaskSource::dispatch
             at /home/cassidy/.cargo/git/checkouts/glib-928cf7b282977403/fe5f2aa/src/main_context_futures.rs:72
  19: g_main_context_dispatch
  20: <unknown>
  21: g_main_loop_run
  22: glib::auto::main_loop::MainLoop::run
             at /home/cassidy/.cargo/git/checkouts/glib-928cf7b282977403/fe5f2aa/src/auto/main_loop.rs:46
  23: glib::main_context_futures::<impl glib::auto::main_context::MainContext>::block_on
             at /home/cassidy/.cargo/git/checkouts/glib-928cf7b282977403/fe5f2aa/src/main_context_futures.rs:302
  24: futures::async_buf_read
             at tests/futures.rs:85
  25: futures::async_buf_read::{{closure}}
             at tests/futures.rs:64
  26: core::ops::function::FnOnce::call_once
             at /rustc/8d69840ab92ea7f4d323420088dd8c9775f180cd/src/libcore/ops/function.rs:232
  27: <alloc::boxed::Box<F> as core::ops::function::FnOnce<A>>::call_once
             at /rustc/8d69840ab92ea7f4d323420088dd8c9775f180cd/src/liballoc/boxed.rs:1017
  28: __rust_maybe_catch_panic
             at src/libpanic_unwind/lib.rs:86
  29: std::panicking::try
             at /rustc/8d69840ab92ea7f4d323420088dd8c9775f180cd/src/libstd/panicking.rs:281
  30: std::panic::catch_unwind
             at /rustc/8d69840ab92ea7f4d323420088dd8c9775f180cd/src/libstd/panic.rs:394
  31: test::run_test_in_process
             at src/libtest/lib.rs:542
  32: test::run_test::run_test_inner::{{closure}}
             at src/libtest/lib.rs:451
```

I guess that's because of `cargo tests` threading but I'm not sure what I'm going wrong.

sdroege:
> I guess that's because of `cargo tests` threading but I'm not sure what I'm going wrong.

You're using the same, global default main context in both threads. So things can happen from the wrong thread now. You need to create a new main context at the beginning of each thread and use that one then.

src/input_stream.rs
```diff
@@ -341,6 +351,221 @@ impl<T: IsA<InputStream> + IsA<Seekable>> io::Seek for InputStreamRead<T> {
...
+        self.state.get_pending()
+    }
+
+    fn get_data(&self) -> Poll<io::Result<&[u8]>> {
+        if let State::HasData {
+            ref buffer,
+            valid: (i, j),
+        } = self.state
+        {
+            return Poll::Ready(Ok(&buffer[i..j + 1]));
```
sdroege, line 448: Why make the second one inclusive if you need to add +1 because of that here?

src/input_stream.rs
```diff
@@ -341,6 +351,221 @@ impl<T: IsA<InputStream> + IsA<Seekable>> io::Seek for InputStreamRead<T> {
...
+
+        if let State::HasData { .. } = self.state {
+            let has_data = mem::replace(&mut self.state, State::Transitioning);
+            if let State::HasData {
+                buffer,
+                valid: (i, j),
+            } = has_data
+            {
+                let available = j - i + 1;
+                let remaining = available - amt;
```
sdroege, line 513: Should this panic if more data is tried to be consumed than is currently available? What do the docs say?

gdesmott:
```
this function may do odd things if poll_fill_buf isn't called before calling it.
The amt must be <= the number of bytes in the buffer returned by poll_fill_buf.
```
So you're on your own if you don't play by the rule. Do you want an explicit panic?

src/input_stream.rs
gdesmott: Indeed, that makes arithmetic simpler.

src/input_stream.rs
sdroege: Sounds like a good idea

gdesmott:
Fixed the tests threading issue as well, thanks for your help.

src/input_stream.rs
gdesmott: Done. All your comments should be fixed now.

PR #299, opened 2020-05-29
Increase reference count of the stored streams in the IOStream when s…

sdroege:
…ubclassing

We return it transfer-none to the caller so our reference would be
dropped at the end of the scope, and then a second time when the
IOStream is finalized. As such we have to increase the reference count
once to prevent a double-free.

----

CC @GuillaumeGomez @EPashkin


EPashkin:
@sdroege Good catch, thanks, 👍

Issue #300, opened 2020-06-02
ThreadedSocketService panics while accepting connection


wikylyu:
I'm learning to use gio. And below is my code

```
use glib::MainContext;
use glib::MainLoop;
use gio::ThreadedSocketService;
use gio::prelude::*;

fn main() {
    let mainloop = MainLoop::new(Some(&MainContext::default()),false);

    let service  = ThreadedSocketService::new(-1);

    if let Err(err) = service.add_inet_port::<glib::Object>(32321, None) {
        return;
    }
    println!("listen 32321");

    service.connect_run(move |_,c,_|->bool{
        println!("{:?}",c);
        return false;
    });

    mainloop.run();
}
```

When ThreadedSocketService receives a connection, it panics.

```
   Compiling gtk v0.1.0 (/home/wiky/temp/rust/gtk)
warning: unused variable: `err`
  --> src/main.rs:11:16
   |
11 |     if let Err(err) = service.add_inet_port::<glib::Object>(32321, None) {
   |                ^^^ help: consider prefixing with an underscore: `_err`
   |
   = note: `#[warn(unused_variables)]` on by default

    Finished dev [unoptimized + debuginfo] target(s) in 1.67s
     Running `target/debug/gtk`
listen 32321
```

***connection established***

```
thread '<unnamed>' panicked at 'assertion failed: crate::types::instance_of::<Self>(ptr as *const _)', <::std::macros::panic macros>:2:4
stack backtrace:
   0:     0x55d6a79929f4 - backtrace::backtrace::libunwind::trace::h5d52ba5f20882f09
                               at /cargo/registry/src/github.com-1ecc6299db9ec823/backtrace-0.3.44/src/backtrace/libunwind.rs:86
   1:     0x55d6a79929f4 - backtrace::backtrace::trace_unsynchronized::hceee092869668a74
                               at /cargo/registry/src/github.com-1ecc6299db9ec823/backtrace-0.3.44/src/backtrace/mod.rs:66
   2:     0x55d6a79929f4 - std::sys_common::backtrace::_print_fmt::ha312c2904605e4d5
                               at src/libstd/sys_common/backtrace.rs:78
   3:     0x55d6a79929f4 - <std::sys_common::backtrace::_print::DisplayBacktrace as core::fmt::Display>::fmt::h5b9981092140b727
                               at src/libstd/sys_common/backtrace.rs:59
   4:     0x55d6a79ab7fc - core::fmt::write::h5f6d7d8de88b4173
[...]

PR #301, opened 2020-06-02
Mark the source_object argument to the ThreadedSocketService::run as …

sdroege:
…nullable

Just like everywhere else.

Fixes https://github.com/gtk-rs/gio/issues/300


sdroege:
CC @EPashkin @GuillaumeGomez

EPashkin:
@sdroege Thanks, 👍

PR #303, opened 2020-06-02
Generate D-Bus classes

myfreeweb:
That was easy enough, even extracting the method invocation object in the `Closure` just worked :)

the "unused import: `glib::translate`" is annoying though

<details>
<summary>Example code</summary>

```rust
    use gio::ApplicationExt;
    let app = gio::Application::new(Some("org.gtk-rs.Test"), gio::ApplicationFlags::default());
    app.register::<gio::Cancellable>(None).unwrap();
    let bus = app.get_dbus_connection().unwrap();

    // Plain call
    let result = bus.call_future(
        Some("org.gnome.Lollypop"), "/org/mpris/MediaPlayer2", "org.mpris.MediaPlayer2.Player",
        "PlayPause", None, None, gio::DBusCallFlags::NONE, 69).await;
    eprintln!("{:?}", result);

    // Proxy
    use gio::DBusProxyExt;
    let proxy = gio::DBusProxy::new_future(
        &bus, gio::DBusProxyFlags::NONE, None,
        Some("org.gnome.Lollypop"), "/org/mpris/MediaPlayer2", "org.mpris.MediaPlayer2.Player"
        ).await.unwrap();
    let result = proxy.call_future(
        "PlayPause", None, gio::DBusCallFlags::NONE, 69).await;
    eprintln!("{:?}", result);

    // Server (v2_46)
    let intf_info = gio::DBusNodeInfo::new_for_xml(r#"
<!DOCTYPE node PUBLIC "-//freedesktop//DTD D-BUS Object Introspection 1.0//EN" "http://www.freedesktop.org/standards/dbus/1.0/introspect.dtd">
<node>
	<interface name="org.gtk_rs.Thingy">
		<method name="Hello">
			<arg name="name" type="s" direction="in"/>
		</method>
	</interface>
</node>
"#).unwrap().lookup_interface("org.gtk_rs.Thingy").unwrap();
    bus.register_object_with_closures(
        "/org/gtk_rs/Thingy", &intf_info,
        Some(&glib::Closure::new(|args| {
            let mut iter = args.iter();
            let _ = iter.next(); // GDBusConnection
            eprintln!("uniq {:?}", iter.next().unwrap().get::<String>());
            eprintln!("path {:?}", iter.next().unwrap().get::<String>());
            eprintln!("intf {:?}", iter.next().unwrap().get::<String>());
            eprintln!("meth {:?}", iter.next().unwrap().get::<String>());
            eprintln!("args {:?}", iter.next().unwrap().get::<glib::Variant>().unwrap().unwrap());
            let invo = iter.next().unwrap().get::<gio::DBusMethodInvocation>().unwrap().unwrap();
            invo.return_value(None);
            None
        })),
        None,
        None).unwrap();
```

</details>

Screen_2020-06-02-16:43:42


src/auto/dbus_connection.rs
```diff
@@ -0,0 +1,1264 @@
...
+            );
+            if error.is_null() {
+                Ok((from_glib_full(ret), from_glib_full(out_fd_list)))
+            } else {
+                Err(from_glib_full(error))
+            }
+        }
+    }
+
+    pub fn close<P: IsA<Cancellable>, Q: FnOnce(Result<(), glib::Error>) + Send + 'static>(
```
GuillaumeGomez, line 417: Is there a mechanism to "open" a connection? If not, I think it should take `self`.

src/auto/dbus_connection.rs
```diff
@@ -0,0 +1,1264 @@
...
+                self.to_glib_none().0,
+                cancellable.map(|p| p.as_ref()).to_glib_none().0,
+                Some(callback),
+                Box_::into_raw(user_data) as *mut _,
+            );
+        }
+    }
+
+    pub fn close_future(
+        &self,
```
GuillaumeGomez, line 453: Same as above.

src/auto/dbus_connection.rs
```diff
@@ -0,0 +1,1264 @@
...
+            let cancellable = Cancellable::new();
+            obj.close(Some(&cancellable), move |res| {
+                send.resolve(res);
+            });
+
+            cancellable
+        }))
+    }
+
+    pub fn close_sync<P: IsA<Cancellable>>(
```
GuillaumeGomez, line 465: I'm curious about this one: how is it different from the `close` method?

src/auto/dbus_connection.rs
```diff
@@ -0,0 +1,1264 @@
...
+            let cancellable = Cancellable::new();
+            obj.flush(Some(&cancellable), move |res| {
+                send.resolve(res);
+            });
+
+            cancellable
+        }))
+    }
+
+    pub fn flush_sync<P: IsA<Cancellable>>(
```
GuillaumeGomez, line 601: How is it different than `flush`?

src/auto/dbus_message.rs
```diff
@@ -0,0 +1,427 @@
...
+    #[cfg(any(unix, feature = "dox"))]
+    pub fn get_unix_fd_list(&self) -> Option<UnixFDList> {
+        unsafe {
+            from_glib_none(gio_sys::g_dbus_message_get_unix_fd_list(
+                self.to_glib_none().0,
+            ))
+        }
+    }
+
+    pub fn lock(&self) {
```
GuillaumeGomez, line 199: Question: what is "lock" used for? If some operations cannot be done if the message is unlocked, then this method should return a wrapper which implements those methods (actually, maybe it should return a wrapper anyway).

src/auto/dbus_connection.rs
myfreeweb: as in "reopen"? no.

Is there a way to make it take `self`? Without manual code I mean >_<

src/auto/dbus_proxy.rs
```diff
@@ -0,0 +1,927 @@
...
+            );
+            if error.is_null() {
+                Ok(from_glib_full(ret))
+            } else {
+                Err(from_glib_full(error))
+            }
+        }
+    }
+
+    pub fn new_sync<P: IsA<Cancellable>>(
```
GuillaumeGomez, line 70: Just a note: I still don't get why there is both `_sync` and the "normal" one. It's weird to see that in the GIO API...

src/auto/dbus_proxy.rs
```diff
@@ -0,0 +1,927 @@
...
+                move |res| {
+                    send.resolve(res);
+                },
+            );
+
+            cancellable
+        }))
+    }
+
+    pub fn new_for_bus<
```
GuillaumeGomez, line 175: Surprisingly, there is no "new_for_bus_sync"...

src/auto/unix_fd_list.rs
```diff
@@ -0,0 +1,121 @@
...
+
+use gio_sys;
+use glib;
+use glib::object::IsA;
+use glib::translate::*;
+use std::fmt;
+use std::mem;
+use std::ptr;
+
+glib_wrapper! {
```
GuillaumeGomez, line 13: This whole struct sounds **very** unsafe. Shouldn't we handle instead types implementing `AsRawFD` instead of `i32`?

src/auto/dbus_connection.rs
GuillaumeGomez: I'm not sure. You can update an argument type but not sure about `self`... Any idea @EPashkin ?

GuillaumeGomez:
That's a great start! I think a lot of small things will need to be updated but at least it has been started, thank you for that!

src/auto/dbus_message.rs
myfreeweb: Apparently locking just makes mutations into no-ops that warn:

https://gitlab.gnome.org/GNOME/glib/-/blob/c93318953eec6a65de3e466f648a4f26109ff0eb/gio/gdbusmessage.c#L835-840

src/auto/functions.rs
```diff
@@ -77,19 +154,19 @@ pub fn bus_unwatch_name(watcher_id: u32) {
...
+//pub fn bus_watch_name(bus_type: BusType, name: &str, flags: /*Ignored*/BusNameWatcherFlags, name_appeared_handler: Option<Box_<dyn Fn(&DBusConnection, &str, &str) + 'static>>, name_vanished_handler: Option<Box_<dyn Fn(&DBusConnection, &str) + 'static>>) -> u32 {
 //    unsafe { TODO: call gio_sys:g_bus_watch_name() }
 //}
 
-//pub fn bus_watch_name_on_connection(connection: /*Ignored*/&DBusConnection, name: &str, flags: /*Ignored*/BusNameWatcherFlags, name_appeared_handler: /*Unimplemented*/Fn(/*Ignored*/DBusConnection, &str, &str), name_vanished_handler: /*Unimplemented*/Fn(/*Ignored*/DBusConnection, &str), user_data: /*Unimplemented*/Option<Fundamental: Pointer>) -> u32 {
+//pub fn bus_watch_name_on_connection(connection: &DBusConnection, name: &str, flags: /*Ignored*/BusNameWatcherFlags, name_appeared_handler: Option<Box_<dyn Fn(&DBusConnection, &str, &str) + 'static>>, name_vanished_handler: Option<Box_<dyn Fn(&DBusConnection, &str) + 'static>>) -> u32 {
```
EPashkin, line 161: Maybe `BusNameWatcherFlags` can generated too?

src/auto/dbus_proxy.rs
myfreeweb: Lots of stuff in this API has sync and async versions, because.. reasons?

("normal" is async)

src/auto/functions.rs
myfreeweb: Sure, nice catch!

src/auto/dbus_proxy.rs
GuillaumeGomez: No I mean, there is "normal", "async" and "not_async". One of the three is very likely unneeded. :)

src/auto/dbus_connection.rs
sdroege: `self` makes no sense here because it's all reference counted. Just like the other `close()` functions we have.

This is ok as is.

src/auto/dbus_connection.rs
sdroege: The non-sync ones are async functions.

src/auto/dbus_connection.rs
sdroege: See above :)

src/auto/dbus_message.rs
sdroege: This should probably be manually implemented similar to `MutexGuard`: have some object returned that on `Drop` unlocks

src/auto/dbus_method_invocation.rs
```diff
@@ -0,0 +1,168 @@
...
+    pub fn return_value(&self, parameters: Option<&glib::Variant>) {
+        unsafe {
+            gio_sys::g_dbus_method_invocation_return_value(
+                self.to_glib_full(),
+                parameters.to_glib_none().0,
+            );
+        }
+    }
+
+    pub fn return_value_with_unix_fd_list<P: IsA<UnixFDList>>(
```
EPashkin, line 149: Need `cfg_condition`

src/auto/dbus_proxy.rs
sdroege: The DBus API has `_sync` (blocking) and without suffix (async function). It's just the opposite of elsewhere for reasons.

src/auto/unix_fd_list.rs
sdroege: Yes this needs to be manually implemented

src/auto/dbus_method_invocation.rs
myfreeweb: yep, I just fixed that in my local copy, after noticing the Appveyor failure

src/auto/unix_fd_list.rs
myfreeweb: `RawFd` is just an alias to `c_int` — would be nice to use it, sure.

I'm not sure how `AsRawFd` offers *safety* advantages though. Convenience, yes — you don't have to convert to the raw fd manually.

The standard library considers `FromRawFd` unsafe, and `libc` functions that do stuff on fds are unsafe, but just passing them around shouldn't be a problem.

src/auto/dbus_arg_info.rs
```diff
@@ -0,0 +1,17 @@
+// This file was generated by gir (https://github.com/gtk-rs/gir)
+// from gir-files (https://github.com/gtk-rs/gir-files)
+// DO NOT EDIT
+
+use gio_sys;
+use glib::translate::*;
+
+glib_wrapper! {
+    #[derive(Debug, PartialEq, Eq, PartialOrd, Ord, Hash)]
+    pub struct DBusArgInfo(Shared<gio_sys::GDBusArgInfo>);
```
sdroege, line 10: This is not useful as it is. Needs some accessors for the struct fields manually implemented, same for `DbusPropertyInfo` and `DbusMethodInfo` and all those other `Info` structs

src/auto/dbus_message.rs
myfreeweb: There is no `unlock`!

This is like a `freeze` thing that makes the actual object behind the ref count immutable and mutations become no-ops.

src/auto/unix_fd_list.rs
GuillaumeGomez: Because that would prevent users to give whatever `i32` they want. We're providing a **safe** API, therefore we can't allow that.

src/auto/dbus_message.rs
GuillaumeGomez: This whole thing makes no sense... :rofl:

sdroege: Ah I see. In that case it should be fine like this. Ugly that it just prints a warning and doesn't return an actual error otherwise but changing that on our side would require to implement all these manually and first check whether the message is locked.

src/auto/dbus_arg_info.rs
myfreeweb: The main use of `Info` structs is to do `gio::DBusNodeInfo::new_for_xml`, `.lookup_interface` and pass the `DBusInterfaceInfo` to `register_object_with_closures` :)

Actual inspection of info would be nice I guess, but not a priority for me.

src/auto/dbus_message.rs
sdroege: > This whole thing makes no sense...

It makes a lot of sense, just that this being C the API is awful. In Rust you would have a builder (or otherwise two different types), and then at some point "build" the message and get an actual (locked, frozen, immutable) message. Here you can modify things and then at some point decide to lock it to make it immutable. Ugly but that's how it is.

src/auto/dbus_arg_info.rs
sdroege: Doesn't look like much is missing for that though, just the field accessors or not?

sdroege: (should create an issue about that so it can be added later and is not forgotten if you don't want to add it)

myfreeweb: Nothing is missing, you can see I'm using it in the example in the first post here.

src/auto/unix_fd_list.rs
myfreeweb: Slightly deter, not prevent ;)

```rust
use std::os::unix::io::{AsRawFd, RawFd};
struct Lol;
impl AsRawFd for Lol {
    fn as_raw_fd(&self) -> RawFd { 420_69 }
}
```

but sure, I'll implement that, for convenience

src/auto/dbus_arg_info.rs
sdroege: No I mean inspection is missing, and the struct accessors for these `Info` structs seem to be the only thing missing for that :)

You're right that it can actually also be used otherwise like you did via the XML, I hadn't thought of that.

sdroege:
Overall looks quite nice, thanks :) It's all rather low-level (like the C API is) but could be wrapped in a nicer higher-level API on top of this for something more convenient.

src/auto/unix_fd_list.rs
myfreeweb: Okay, now using `IntoRawFd` for `g_unix_fd_list_new_from_array` (since "file descriptors become the property of the new list and may no longer be used by the caller") and `RawFd` for the rest.

src/auto/dbus_arg_info.rs
sdroege: I'll add these once this one is merged here

src/auto/dbus_interface.rs
```diff
@@ -0,0 +1,71 @@
...
+
+    fn get_info(&self) -> Option<DBusInterfaceInfo>;
+
+    fn get_object(&self) -> Option<DBusObject>;
+
+    fn set_object<P: IsA<DBusObject>>(&self, object: Option<&P>);
+}
+
+impl<O: IsA<DBusInterface>> DBusInterfaceExt for O {
+    fn dup_object(&self) -> Option<DBusObject> {
```
sdroege, line 33: `dup_object()` should be renamed to `get_object()`, and `get_object()` be hidden. `get_object()` is not safe, see the docs, and otherwise equivalent.

src/auto/dbus_interface_info.rs
```diff
@@ -0,0 +1,72 @@
...
+}
+
+impl DBusInterfaceInfo {
+    pub fn cache_build(&self) {
+        unsafe {
+            gio_sys::g_dbus_interface_info_cache_build(self.to_glib_none().0);
+        }
+    }
+
+    pub fn cache_release(&self) {
```
sdroege, line 30: `cache_build()` and `cache_release()` should probably get a return type that calls the latter on `Drop`

src/auto/dbus_message.rs
```diff
@@ -0,0 +1,427 @@
...
+
+    pub fn new_method_reply(&self) -> Option<DBusMessage> {
+        unsafe {
+            from_glib_full(gio_sys::g_dbus_message_new_method_reply(
+                self.to_glib_none().0,
+            ))
+        }
+    }
+
+    pub fn print(&self, indent: u32) -> Option<GString> {
```
sdroege, line 235: This could become the `Display` impl :)

src/auto/dbus_method_invocation.rs
```diff
@@ -0,0 +1,171 @@
...
+                error_message.to_glib_none().0,
+            );
+        }
+    }
+
+    //pub fn return_error(&self, domain: glib::Quark, code: i32, format: &str, : /*Unknown conversion*//*Unimplemented*/Fundamental: VarArgs) {
+    //    unsafe { TODO: call gio_sys:g_dbus_method_invocation_return_error() }
+    //}
+
+    pub fn return_error_literal(&self, domain: glib::Quark, code: i32, message: &str) {
```
sdroege, line 118:
```suggestion
    pub fn return_error<T: glib::ErrorDomain>(&self, error: T, message: &str) {
```

Also if there are any other similar functions :)

src/auto/dbus_method_invocation.rs
```diff
@@ -0,0 +1,171 @@
...
+                message.to_glib_none().0,
+            );
+        }
+    }
+
+    //pub fn return_error_valist(&self, domain: glib::Quark, code: i32, format: &str, var_args: /*Unknown conversion*//*Unimplemented*/Unsupported) {
+    //    unsafe { TODO: call gio_sys:g_dbus_method_invocation_return_error_valist() }
+    //}
+
+    pub fn return_gerror(&self, error: &glib::Error) {
```
sdroege, line 133:
```suggestion
    pub fn return_gerror(&self, error: glib::Error) {
```

src/auto/dbus_interface_info.rs
myfreeweb: It's a ref counted type of thing..

> If this has already been called with info , the existing cache is used and its use count is increased.

> Decrements the usage count for the cache for info built by g_dbus_interface_info_cache_build() (if any) and frees the resources used by the cache if the usage count drops to zero.

except an internal thing that you can't touch, so I don't think we need a return type

src/auto/functions.rs
```diff
@@ -24,46 +28,120 @@ use Resource;
...
 
-//pub fn bus_own_name_on_connection(connection: /*Ignored*/&DBusConnection, name: &str, flags: /*Ignored*/BusNameOwnerFlags, name_acquired_handler: /*Unimplemented*/Fn(/*Ignored*/DBusConnection, &str), name_lost_handler: /*Unimplemented*/Fn(/*Ignored*/DBusConnection, &str), user_data: /*Unimplemented*/Option<Fundamental: Pointer>) -> u32 {
+//pub fn bus_own_name_on_connection(connection: &DBusConnection, name: &str, flags: BusNameOwnerFlags, name_acquired_handler: Option<Box_<dyn Fn(&DBusConnection, &str) + 'static>>, name_lost_handler: Option<Box_<dyn Fn(&DBusConnection, &str) + 'static>>) -> u32 {
 //    unsafe { TODO: call gio_sys:g_bus_own_name_on_connection() }
 //}
 
-//pub fn bus_own_name_on_connection_with_closures(connection: /*Ignored*/&DBusConnection, name: &str, flags: /*Ignored*/BusNameOwnerFlags, name_acquired_closure: /*Ignored*/Option<&glib::Closure>, name_lost_closure: /*Ignored*/Option<&glib::Closure>) -> u32 {
-//    unsafe { TODO: call gio_sys:g_bus_own_name_on_connection_with_closures() }
-//}
+pub fn bus_own_name_on_connection_with_closures(
```
sdroege, line 108: This should be a) renamed to not include the "with_closures" part and b) the closures have fixed signatures so it would be better for users of this API to instead take proper Rust closures with the correct types so users don't have to work with `glib::Value`s :)

src/auto/functions.rs
```diff
@@ -24,46 +28,120 @@ use Resource;
...
+            name_acquired_closure.to_glib_none().0,
+            name_lost_closure.to_glib_none().0,
+        )
+    }
+}
 
-//pub fn bus_own_name_with_closures(bus_type: /*Ignored*/BusType, name: &str, flags: /*Ignored*/BusNameOwnerFlags, bus_acquired_closure: /*Ignored*/Option<&glib::Closure>, name_acquired_closure: /*Ignored*/Option<&glib::Closure>, name_lost_closure: /*Ignored*/Option<&glib::Closure>) -> u32 {
-//    unsafe { TODO: call gio_sys:g_bus_own_name_with_closures() }
-//}
+pub fn bus_own_name_with_closures(
```
sdroege, line 126: Same as for `bus_own_name_on_connection_with_closures`

src/auto/functions.rs
```diff
@@ -77,21 +155,49 @@ pub fn bus_unwatch_name(watcher_id: u32) {
...
 
-//pub fn bus_watch_name_on_connection(connection: /*Ignored*/&DBusConnection, name: &str, flags: /*Ignored*/BusNameWatcherFlags, name_appeared_handler: /*Unimplemented*/Fn(/*Ignored*/DBusConnection, &str, &str), name_vanished_handler: /*Unimplemented*/Fn(/*Ignored*/DBusConnection, &str), user_data: /*Unimplemented*/Option<Fundamental: Pointer>) -> u32 {
+//pub fn bus_watch_name_on_connection(connection: &DBusConnection, name: &str, flags: BusNameWatcherFlags, name_appeared_handler: Option<Box_<dyn Fn(&DBusConnection, &str, &str) + 'static>>, name_vanished_handler: Option<Box_<dyn Fn(&DBusConnection, &str) + 'static>>) -> u32 {
 //    unsafe { TODO: call gio_sys:g_bus_watch_name_on_connection() }
 //}
 
-//pub fn bus_watch_name_on_connection_with_closures(connection: /*Ignored*/&DBusConnection, name: &str, flags: /*Ignored*/BusNameWatcherFlags, name_appeared_closure: /*Ignored*/Option<&glib::Closure>, name_vanished_closure: /*Ignored*/Option<&glib::Closure>) -> u32 {
-//    unsafe { TODO: call gio_sys:g_bus_watch_name_on_connection_with_closures() }
-//}
+pub fn bus_watch_name_on_connection_with_closures(
```
sdroege, line 166: Same as for `bus_own_name_on_connection_with_closures`

src/auto/functions.rs
```diff
@@ -77,21 +155,49 @@ pub fn bus_unwatch_name(watcher_id: u32) {
...
+            name_appeared_closure.to_glib_none().0,
+            name_vanished_closure.to_glib_none().0,
+        )
+    }
+}
 
-//pub fn bus_watch_name_with_closures(bus_type: /*Ignored*/BusType, name: &str, flags: /*Ignored*/BusNameWatcherFlags, name_appeared_closure: /*Ignored*/Option<&glib::Closure>, name_vanished_closure: /*Ignored*/Option<&glib::Closure>) -> u32 {
-//    unsafe { TODO: call gio_sys:g_bus_watch_name_with_closures() }
-//}
+pub fn bus_watch_name_with_closures(
```
sdroege, line 184: Same as for `bus_own_name_on_connection_with_closures`

src/auto/dbus_connection.rs
```diff
@@ -0,0 +1,1264 @@
...
+    pub fn is_closed(&self) -> bool {
+        unsafe { from_glib(gio_sys::g_dbus_connection_is_closed(self.to_glib_none().0)) }
+    }
+
+    //pub fn register_object(&self, object_path: &str, interface_info: &DBusInterfaceInfo, vtable: /*Ignored*/Option<&DBusInterfaceVTable>, user_data: /*Unimplemented*/Option<Fundamental: Pointer>) -> Result<(), glib::Error> {
+    //    unsafe { TODO: call gio_sys:g_dbus_connection_register_object() }
+    //}
+
+    #[cfg(any(feature = "v2_46", feature = "dox"))]
+    pub fn register_object_with_closures(
```
sdroege, line 678: Same as for `bus_own_name_on_connection_with_closures`

src/auto/dbus_connection.rs
```diff
@@ -0,0 +1,1264 @@
...
+            );
+            if error.is_null() {
+                Ok(())
+            } else {
+                Err(from_glib_full(error))
+            }
+        }
+    }
+
+    //pub fn register_subtree(&self, object_path: &str, vtable: /*Ignored*/&DBusSubtreeVTable, flags: /*Ignored*/DBusSubtreeFlags, user_data: /*Unimplemented*/Option<Fundamental: Pointer>) -> Result<(), glib::Error> {
```
sdroege, line 705: Curious that there is no closures variant of this one... https://gitlab.gnome.org/GNOME/glib/-/issues/2122

src/auto/dbus_connection.rs
```diff
@@ -0,0 +1,1264 @@
...
+                arg0.to_glib_none().0,
+                flags.to_glib(),
+                callback,
+                Box_::into_raw(super_callback0) as *mut _,
+                destroy_call9,
+            )
+        }
+    }
+
+    pub fn signal_unsubscribe(&self, subscription_id: u32) {
```
sdroege, line 917: This should get a newtype wrapper, `struct SignalSubscriptionId(u32)`. Similar to `Object::connect()` etc

src/auto/dbus_connection.rs
```diff
@@ -0,0 +1,1264 @@
...
+        }
+    }
+
+    pub fn unexport_menu_model(&self, export_id: u32) {
+        unsafe {
+            gio_sys::g_dbus_connection_unexport_menu_model(self.to_glib_none().0, export_id);
+        }
+    }
+
+    pub fn unregister_object(&self, registration_id: u32) -> bool {
```
sdroege, line 941: This should get a newtype wrapper. Similar to `Object::connect()` etc (of course also for the register function the same)

src/auto/dbus_connection.rs
```diff
@@ -0,0 +1,1264 @@
...
+        unsafe {
+            from_glib(gio_sys::g_dbus_connection_unregister_object(
+                self.to_glib_none().0,
+                registration_id,
+            ))
+        }
+    }
+
+    pub fn unregister_subtree(&self, registration_id: u32) -> bool {
```
sdroege, line 950: This should get a newtype wrapper. Similar to `Object::connect()` etc

src/auto/dbus_connection.rs
```diff
@@ -0,0 +1,1264 @@
...
+        unsafe {
```
sdroege, line 951: Also all the unregister functions returning bools should return a `Result<(), glib::BoolError>`. `false` is an actual error

sdroege:
Proper use of this API will also need https://github.com/gtk-rs/glib/issues/113 to be solved so more than trivial types can be passed via method calls, etc.

src/auto/dbus_interface_info.rs
sdroege: The problem is that without having a such a type, you would "forget" to decrement it on panicking and probably also accidentally in functions with multiple return paths.

Each `cache_build()` would return a value (refcount +1), each `Drop` of them would call `cache_release() (refcount -1).

Am I missing something?

src/auto/dbus_connection.rs
EPashkin: As currently no way to register subtree, not better to just ignore this function?

myfreeweb: Yes, I'll ignore the subtree stuff

src/auto/dbus_connection.rs
EPashkin: Note that we have flag for that:
https://github.com/gtk-rs/gir/blob/master/README.md
```
        # convert bool return types to Result<(), glib::BoolError> with
        # the given error message on failure
        bool_return_is_error = "Function failed doing what it is supposed to do"
```

src/auto/dbus_interface_info.rs
myfreeweb: Yeah I guess that's nice. I don't care about this API though, maybe I'll just ignore cache_ methods for now?

sdroege: Yes please, and afterwards we should create an issue with a list of all the missing things (like also the inspection and subtree APIs).

src/auto/dbus_interface.rs
myfreeweb: tbh I'm not sure about the "rename" part. Don't like it when bindings are that unfaithful to the original library :D

sdroege: From a bindings point of view both functions work exactly the same, and both are "dup"licating (i.e. `g_object_ref()`) the return value. Just that the `get()` variant is internally unsafe while the `dup()` one is not (the `get()` one was a API design bug from what I understand).

So we definitely don't want to expose the `get()` variant, and as `dup()` makes semantically no sense for the bindings renaming it to `get()` would seem better and less confusing for users :)

src/auto/dbus_connection.rs
myfreeweb: interestingly, the generator decided to ignore the returned ID altogether..

sdroege: Hrm... @EPashkin any guesses why? :)

EPashkin: Result of very old https://github.com/gtk-rs/gir/pull/187
Can overridden with
```toml
[[object.function]]
name = "register_object_with_closures"
[object.function.return]
use_return_for_result = true
```

myfreeweb:
Added newtypes for everything in connection. Absolutely hated this process, editing generated code is evil >_< but whatever, not my project haha.

Also the closures are typed now. `name_lost` has a nullable connection because it can be called when connecting to the bus failed: https://developer.gnome.org/gio/stable/gio-Owning-Bus-Names.html#g-bus-own-name

src/dbus_connection.rs
```diff
@@ -0,0 +1,355 @@
...
+                        value,
+                    );
+                    Some(result.to_value())
+                })
+                .to_glib_none()
+                .0,
+                &mut error,
+            );
+            if error.is_null() {
+                Ok(RegistrationId(NonZeroU32::new_unchecked(id)))
```
EPashkin, line 149: Maybe safer to use `new(id).expect' ?

myfreeweb:
And closures themselves are not optional, because you'd have to provide a type argument anyway..

src/dbus_connection.rs
myfreeweb: This is what is used in `glib`.

I don't think we really need to test gio's function contracts on every call :)

EPashkin: `glib` checked it before use `new_unchecked`

sdroege: Yeah either add an assertion before and `new_unchecked()`, or `new()` and `expect()`. glib didn't do the latter because in older Rust you don't get too useful source file / line number with `expect()`.

sdroege:
> Added newtypes for everything in connection. Absolutely hated this process, editing generated code is evil >_< but whatever, not my project haha.

Yeah it sucks but I'm not sure what the alternative would be other than keeping the API suboptimal. The goal is not to keep the amount of work we have to do low :) There are unfortunately various patterns that we can't detect automatically.

I'll go through the whole code again later today but I guess this is basically read now :)

Gir.toml
```diff
@@ -359,6 +402,100 @@ status = "generate"
...
+    #wrapped with newtype manually
+    ignore = true
+
+[[object]]
+name = "Gio.DBusInterface"
+status = "generate"
+    [[object.function]]
+    name = "get_object"
+    # returns non-thread-safe reference
+    ignore = true
```
sdroege, line 446: And with `rename = "get"` for the `dup_object` function please :)

src/dbus.rs
```diff
@@ -0,0 +1,166 @@
...
+
+use gio_sys;
+use glib::translate::*;
+use std::num::NonZeroU32;
+use BusNameOwnerFlags;
+use BusNameWatcherFlags;
+use BusType;
+use DBusConnection;
+
+pub struct OwnerId(NonZeroU32);
```
sdroege, line 13: `#[derive(Debug, Eq, PartialEq)]`

Also for all the other ones

src/dbus.rs
```diff
@@ -0,0 +1,166 @@
...
+pub struct WatcherId(NonZeroU32);
+
+fn own_closure<F>(f: F) -> glib::Closure
+where
+    F: Fn(DBusConnection, &str) + Send + Sync + 'static,
+{
+    glib::Closure::new(move |args| {
+        let mut iter = args.iter();
```
sdroege, line 21: I would usually just do
```rust
let conn = args[0].get::<DbusConnection>().unwrap().unwrap();
let name = args[1].get::<&str>().unwrap().unwrap();
```

Seems less noisy than the iterator

src/dbus.rs
```diff
@@ -0,0 +1,166 @@
...
+        None
+    })
+}
+
+pub fn bus_own_name_on_connection<NameAcquired, NameLost>(
+    connection: &DBusConnection,
+    name: &str,
+    flags: BusNameOwnerFlags,
+    name_acquired_closure: NameAcquired,
+    name_lost_closure: NameLost,
```
sdroege, line 58: I would remove the `_closure` suffix here (and elsewhere), just seems a bit noisy :)

myfreeweb:
Done.

> I'm not sure what the alternative would be

Adding more and more features to `Gir.toml` all the time? ;)

EPashkin:
IMHO it looks good, Thanks, @myfreeweb

sdroege:
@myfreeweb Do you want to provide your example from the screenshot or something similar (like using the notification dbus interface) as an example to https://github.com/gtk-rs/examples/ ?

myfreeweb:
Yes, sure. (How are versions handled? I see the examples repo doesn't use git versions of crates..)

GuillaumeGomez:
Look at the pending branch. ;)

PR #305, opened 2020-06-08
Focal

EPashkin:
Part of gtk-rs/gir-files#55


src/auto/constants.rs
```diff
@@ -514,6 +514,13 @@ pub static FILE_ATTRIBUTE_UNIX_UID: once_cell::sync::Lazy<&'static str> =
             .to_str()
             .unwrap()
     });
+#[cfg(any(feature = "v2_64", feature = "dox"))]
+pub static MEMORY_MONITOR_EXTENSION_POINT_NAME: once_cell::sync::Lazy<&'static str> =
```
sdroege, line 518: You could enable the memory monitor API now, that should all be autogeneratable

EPashkin: Added `MemoryMonitor`

sdroege:
```
error[E0425]: cannot find function `g_get_os_info` in crate `glib_sys`
   --> C:\Users\appveyor\.cargo\git\checkouts\glib-928cf7b282977403\13338dc\src\auto\functions.rs:635:39
    |
635 |     unsafe { from_glib_full(glib_sys::g_get_os_info(key_name.to_glib_none().0)) }
    |                                       ^^^^^^^^^^^^^ not found in `glib_sys`
error: aborting due to previous error
```

Ugh... on Windows. There's a Windows implementation of that function now in GLib master so marking this UNIX-only is also a bit problematic... maybe on Windows we should simply have a trivial fallback implementation on our side?

EPashkin:
sys just not merged, I have this function in `libglib-2.0.a`

sdroege:
> Ugh... on Windows. There's a Windows implementation of that function now in GLib master so marking this UNIX-only is also a bit problematic... maybe on Windows we should simply have a trivial fallback implementation on our side?

Ah because it's building against the old -sys still, because that one was not merged yet. Not a problem.

PR #307, opened 2020-07-03
Regen

GuillaumeGomez:
Linked to gtk-rs/gtk#1028.


GuillaumeGomez:
Why are there so many removals here? O.o

GuillaumeGomez:
Problem was that gio didn't updated its way of declaring builders. :)

PR #315, opened 2020-08-18
Inet addr

### LINKED ISSUE #314: InetAddress <-> standard library IpAddr* types conversion
There isn't a nice way to convert between the `gio::InetAddress` type and the IP address types provided by the standard library ([`Ipv4Addr`](https://doc.rust-lang.org/std/net/struct.Ipv4Addr.html), [`Ipv6Addr`](https://doc.rust-lang.org/std/net/struct.Ipv6Addr.html), as well as the enum abstracting over both, [`IpAddr`](https://doc.rust-lang.org/std/net/struct.IpAddr.html)).

This makes it difficult to use e.g. the gio resolver with the rest of the Rust ecosystem.

GuillaumeGomez:
Fixes #314.

cc @sdroege @EPashkin @elomatreb

### DIFF
--- a/src/inet_address.rs
+++ b/src/inet_address.rs
@@ -6,7 +6,6 @@ use InetAddressExt;
 use SocketFamily;
 
 use std::net::{IpAddr, Ipv4Addr, Ipv6Addr};
-use std::slice::from_raw_parts;
 
 #[derive(Debug)]
 pub enum InetAddressBytes<'a> {
@@ -58,6 +57,31 @@ impl<O: IsA<InetAddress>> InetAddressExtManual for O {
                 Some(InetAddressBytes::V6(&*(bytes as *const [u8; 16])))
             } else {
                 None
+            }
+        }
+    }
+}
+
+impl From<IpAddr> for InetAddress {
+    fn from(addr: IpAddr) -> Self {
+        match addr {
+            IpAddr::V4(v4) => InetAddress::from_bytes(InetAddressBytes::V4(&v4.octets())),
+            IpAddr::V6(v6) => InetAddress::from_bytes(InetAddressBytes::V6(&v6.octets())),
+        }
+    }
+}
+
+impl Into<IpAddr> for InetAddress {
+    fn into(self) -> IpAddr {
+        let size = self.get_native_size();
+        unsafe {
+            let bytes = gio_sys::g_inet_address_to_bytes(self.to_glib_none().0);
+            if size == 4 {
+                IpAddr::V4(Ipv4Addr::from(*(bytes as *const [u8; 4])))
+            } else if size == 16 {
+                IpAddr::V6(Ipv6Addr::from(*(bytes as *const [u16; 8])))
+            } else {
+                panic!("Unknown IP kind");
             }
         }
     }
--- a/src/inet_address.rs
+++ b/src/inet_address.rs
@@ -1,7 +1,12 @@
 use gio_sys;
+use glib::object::IsA;
 use glib::translate::*;
 use InetAddress;
+use InetAddressExt;
 use SocketFamily;
+
+use std::net::{IpAddr, Ipv4Addr, Ipv6Addr};
+use std::slice::from_raw_parts;
 
 #[derive(Debug)]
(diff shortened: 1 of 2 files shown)


src/inet_address.rs
```diff
@@ -38,3 +42,63 @@ impl InetAddress {
...
+    fn to_bytes<'a>(&'a self) -> Option<InetAddressBytes<'a>> {
+        let size = self.get_native_size();
+        unsafe {
+            let bytes = gio_sys::g_inet_address_to_bytes(self.as_ref().to_glib_none().0);
+            if size == 4 {
+                // The `try_into` is used to convert from `&[u8]` to `&[u8; 4]`.
+                Some(InetAddressBytes::V4(
+                    from_raw_parts(bytes, 4)
+                        .try_into()
+                        .expect("failed to convert to array"),
```
sdroege, line 61:
```suggestion
                Some(InetAddressBytes::V4(&*(bytes as *const [u8; 4]),
```

Should work equally well and less line noise. If it doesn't, how does it fail?

src/inet_address.rs
```diff
@@ -38,3 +42,63 @@ impl InetAddress {
...
+}
+
+impl Into<IpAddr> for InetAddress {
+    fn into(self) -> IpAddr {
+        let size = self.get_native_size();
+        unsafe {
+            let bytes = gio_sys::g_inet_address_to_bytes(self.to_glib_none().0);
+            if size == 4 {
+                let bytes = from_raw_parts(bytes, 4);
+                IpAddr::V4(Ipv4Addr::new(bytes[0], bytes[1], bytes[2], bytes[3]))
```
sdroege, line 93: Better via https://doc.rust-lang.org/std/net/struct.Ipv4Addr.html#impl-From%3C%5Bu8%3B%204%5D%3E

```suggestion
                IpAddr::V4(Ipv4Addr::from(bytes))
```

src/inet_address.rs
GuillaumeGomez: Aren't you creating a temporary reference by doing so? We're moving a variable here, not just using a function, so I have a doubt...

sdroege: In what sense? `*const [u8; 4]` is equivalent to `&[u8;4]` and either of those are equivalent to `*const u8` just that there is static type system information about the length.

sdroege: That's how the `try_into()` you're using here works internally too, just that we can skip all these redundant steps (all `try_into()` is checking if the size is 4, which you made sure by passing 4 to `from_raw_parts()`)

GuillaumeGomez: I'm definitely not up-to-date about references/pointers cast...

sdroege: https://doc.rust-lang.org/src/core/array/mod.rs.html#157-163

GuillaumeGomez:
Updated!

src/inet_address.rs
```diff
@@ -38,3 +40,49 @@ impl InetAddress {
...
+impl From<IpAddr> for InetAddress {
+    fn from(addr: IpAddr) -> Self {
+        match addr {
+            IpAddr::V4(v4) => InetAddress::from_bytes(InetAddressBytes::V4(&v4.octets())),
+            IpAddr::V6(v6) => InetAddress::from_bytes(InetAddressBytes::V6(&v6.octets())),
+        }
+    }
+}
+
+impl Into<IpAddr> for InetAddress {
```
elomatreb, line 74: Shouldn't this rather be a fallible conversion, instead of panicking? Also, is relying on the returned size like that safe when from the looks of it the `InetAddress` type can also represent unix paths that don't have a fixed length?

sdroege: This can't really fail though, `InetAddress` / `IpAddr` is either an IPv4 or IPv6 address and nothing else.

UNIX addresses are `GUnixSocketAddress`

elomatreb: Actually nevermind the second part, I got confused by the documentation.

elomatreb:
Thanks a lot for the work!

PR #316, opened 2020-08-27
0.9.1 release

sdroege:
CC @EPashkin @GuillaumeGomez


GuillaumeGomez (approved):

sdroege:
Not sure why this fails like that, but maybe we should add `// rustdoc-stripper-ignore-next` here? That's what it's supposed to be used for, right? @GuillaumeGomez

GuillaumeGomez:
Indeed but it's supposed to put this documentation back. It's weird...

sdroege:
> Indeed but it's supposed to put this documentation back. It's weird...

From where would it put it back? Is it in `lgpl-docs` `vendor.md`? If so we need to update that for this bugfix release, or just remove the docs, or rustdoc-stripper-ignore-next

GuillaumeGomez:
Let's use `rustdoc-stripper-ignore-next` for now.

sdroege:
Ok, seems this works :)

PR #318, opened 2020-10-09
groovy update: regenerate

bilelmoussaoui:


src/auto/tls_connection.rs
```diff
@@ -47,6 +47,9 @@ pub trait TlsConnectionExt: 'static {
 
     fn get_certificate(&self) -> Option<TlsCertificate>;
 
+    //#[cfg(any(feature = "v2_66", feature = "dox"))]
+    //fn get_channel_binding_data(&self, type_: /*Ignored*/TlsChannelBindingType, data: /*Ignored*/glib::ByteArray) -> Result<(), glib::Error>;
```
sdroege, line 51: That can be easily generated if you add the enum to the `generate` list and `GLib.ByteArray` to the `manual` list in `Gir.toml`.

sdroege (commented):
Looks good to me otherwise

src/auto/tls_connection.rs
bilelmoussaoui: Done, adding GLib.ByteArray generated a bunch of other bindings. We should probably check the not bound ones & that could b generated in the future.

sdroege (approved):

sdroege:
@bilelmoussaoui Please regenerate with the merged version of `gir-files` and then let's get it in :)

sdroege (approved):

sdroege:
@GuillaumeGomez The CI failure comes from something docs related. Any idea?

```diff
diff --git b/src/inet_address.rs a/src/inet_address.rs
index 1e9bfeb..0bfa7c8 100644
--- b/src/inet_address.rs
+++ a/src/inet_address.rs
@@ -46,6 +46,7 @@ pub trait InetAddressExtManual {
 }
 
 impl<O: IsA<InetAddress>> InetAddressExtManual for O {
+    /// Returns `None` in case the address has a native size different than 4 and 16.
     fn to_bytes<'a>(&'a self) -> Option<InetAddressBytes<'a>> {
         let size = self.get_native_size();
         unsafe {
The command "git diff -R --exit-code" exited with 1.
```

GuillaumeGomez:
I guess it's a new bug... I'll open an issue on rustdoc-stripper.

sdroege:
Merging then as this is code that is not even touched in this PR :)

PR #322, opened 2020-10-19
Change modification time of FileInfo to work based on std::time::Syst…

### LINKED ISSUE #320: `TimeVal` usage is wrong and maybe shouldn't be there at all
```
src/auto/file_info.rs
213:    pub fn get_modification_time(&self) -> glib::TimeVal {
457:    pub fn set_modification_time(&self, mtime: &mut glib::TimeVal) {
```

The second shouldn't be `mut`, and maybe those should all actually become `std::time::SystemTime` instead.

See also https://github.com/gtk-rs/gdk-pixbuf/pull/158/files#r504535569

sdroege:
…emTime

Fixes https://github.com/gtk-rs/gio/issues/320

### DIFF
--- a/Gir.toml
+++ b/Gir.toml
@@ -59,7 +59,6 @@ generate = [
     "Gio.FileCopyFlags",
     "Gio.FileCreateFlags",
     "Gio.FileIcon",
-    "Gio.FileInfo",
     "Gio.FileInputStream",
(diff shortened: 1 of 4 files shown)


GuillaumeGomez (approved):
Thanks!

sdroege:
CI failures again unrelated.

